PR #3, opened 2015-03-06
Added BOARD_PATH variable so that the full board path can be reached in ...

rhempel:
...the hostfs


dlech:
why do we need a separate variable? BOARD is already a path.

rhempel:
We don't need a separate variable - I just like the symmetry with SCRIPT_PATH - but your're right there's less to break if I make BOARD an absolute path - give me a few minutes

PR #11, opened 2015-12-13
Pull request for "better" awk fix

BrendanSimon:


dlech:
Thanks! I think this is really the best way to do this.

BrendanSimon:
Cool :)
BTW, I didn't update the other boards to remove the perl script.

dlech:
Yeah, I reverted that commit, so it is taken care of.

PR #12, opened 2015-12-25
Robustness improvements

cmacq2:
Robustness improvements: unsharing on Debian, comments in package files, and report hook.

This change adds a few convenience features to improve overall robustness of the brickstrap process:
 - Introduce unshare-harder variant which will prod the kernel into allowing ordinary users to unshare, which is necessary for user-unshare.pl to work on recent-ish Debian.
   Any changes to original kernel settings are reverted on exit. This avoids having to run the entire brickstrap with sudo/su, only the initial prodding and reverting need sudo/su privileges.
 - Add automatic filtering of comments in package files when $PACKAGES is assembled. This simplifies creating derivative board directories for new users (like me).
   Comment lines (starting with ; or #) and empty lines in package files are now ignored instead of being appended to $PACKAGES.
 - Add support for runnning a reporting hook. This permits the user to include a custom script which will be run at the end of sequences executed by 'run-hooks' or 'all'.
   Such a reporting hook is useful to aggregate and report basic information about the rootfs in one single, convenient location of the brickstrap output.
   For example: hostname, usernames, passwords, key signatures... This makes it easier for consumers of the rootfs to get started using their new OS as an opaque binary image.


brickstrap-unshare-harder.sh
```diff
@@ -0,0 +1,25 @@
+#!/bin/sh
```
dlech: I'm not sure how I feel about this. For starters, this is targeted at Debian because Debian doesn't enable `unprivileged_userns_clone` by default. But, Debian also does not use `sudo` by default, so this still won't work by default.

Also, couldn't this just be included in the main brickstrap.sh script?

I would be happy just adding a check for this to the main script and provide instructions on how to fix it manually, similar to checking that the kernel image is readable that is already done.

brickstrap.sh
```diff
@@ -404,7 +440,10 @@ case "${cmd}" in
     copy-root)           copy-root;;
     configure-packages)  configure-packages;;
     run-hook)            run-hook ${BOARDDIR}/hooks/${run_hook_arg};;
-    run-hooks)           run-hooks;;
+    run-hooks)
+        run-hooks
+        dump-info-hook
```
dlech: Since this is run at the same time as other hooks, why not just use a hook for this? I'm actually already doing something like this to generate a changelog template.

We could make sure hook files run in a deterministic order based on filename so that the `zzz-dump-info` would be guaranteed to run last. Of course we could add numeric prefixes to all of the files for better style, then we could have `99-dump-info`.

brickstrap.sh
```diff
@@ -416,6 +455,7 @@ case "${cmd}" in
         create-rootfs
         create-tar
         create-image
+        dump-info-hook
```
dlech: This will make `dump-info-hook` run twice since it is called in `run-hooks` as well.

brickstrap.sh
```diff
@@ -60,7 +60,7 @@ Options
                   0=fail, ... 3=info(default), 4=debug
 
   DEBIAN_MIRROR   Specifies the debian mirror used by apt
-                  default: http://ftp.debian.org/debian
+                  default: http://httpredir.debian.org/debian
```
dlech: look good.

brickstrap.sh
```diff
@@ -198,10 +198,33 @@ set -o pipefail
 #####################################################################
 
 function create-conf() {
+    #
+    # Set the defaults for mirrors as promised by help, if these haven't been configured for the board yet.
```
dlech: looks good.

brickstrap.sh
```diff
@@ -198,10 +198,33 @@ set -o pipefail
...
+        RASPBIAN_MIRROR="http://archive.raspbian.org/raspbian"
+    fi
+    if [ -z "${EV3DEV_MIRROR}" ]; then
+        EV3DEV_MIRROR="http://ev3dev.org/debian"
+    fi
     info "Creating multistrap configuration file..."
     debug "BOARDDIR: ${BOARDDIR}"
     for f in ${BOARDDIR}/packages/*; do
-        while read line; do PACKAGES="${PACKAGES} $line"; done < "$f"
+        while read line; do
```
dlech: One other thing I would like to fix with this is that if there is not a newline at the end of the file, the last line is ignored.

brickstrap.sh
cmacq2: > Since this is run at the same time as other hooks, why not just use a hook for this?

Because it is fundamentally not meant to be the same kind of hook: i.e. the script is not supposed to do arbitrary things but just log relevant info. So it makes more sense (to me) to  simply defer running of the hook to the right point in time, which brickstrap knows, and therefore guarantee a well-defined execution environment.

Also, the way the hooks are currently set up simply naming this specially will not work:
 * The hooks are not sorted by name before being executed in-order. The order will therefore be (on Linux) whatever `*` expansion returns them in. So not stable/predictable. Therefore it would be fragile.
 * Additionally fixing this is not easy to do cleanly because by design the hooks expect to be sourced (so simply executing them through run-parts won't work).

Beyond that hooks run before image creation, and if you wish to log things about the image (e.g. SHA256 sums) you need to run... after.

Deferring the hook until when it is actually meant to be invoked solves all this by design.

brickstrap.sh
cmacq2: No. The way I set it up, it's only called from the case/esac block which invokes the command sequences. So it's called either when you invoke brickstrap with the run-hooks command or with the all command but never twice in the same brickstrap command.

brickstrap.sh
cmacq2: So would I (I just had quite a puzzle as to why that is). I can tinker a bit further with this, but basically it would involve appending (`echo ""`) an additional line and abusing `cat`. Something like: 
```
orig=$(cat $f); 
sanity=$(cat <<<EOF
$(orig)

EOF
);
echo "$(contents)" | while read...
```

Perhaps?

brickstrap-unshare-harder.sh
cmacq2: Those concerns about default Debian is why I kept it separate. It's meant as a fallback for the user to be able to recover easily if they're willing to accept a package and the 'fiddling'. 

The idea is the user tries brickstrap.sh: it works great; it doesn't they can use this as a fallback. Also with this it is simpler to ensure correctness in a sense, because I don't know exactly what else brickstrap.sh needs to trap and if there are edge-cases (which are easy to fix in this wrapper script if they appear, perhaps not so much in brickstrap.sh)...

brickstrap.sh
dlech: Makes more sense to me too. :wink:

Although we don't want it tacked onto the `run-hooks` command. Like the example that you gave, if you want to do a checksum on the image file, the image file has not been created at this point. Instead, I think it should be a new stand-alone command (and be the final step of the `all` command). And maybe have a more generic name.

brickstrap.sh
dlech: Of course, you are right.

brickstrap.sh
dlech: How about one of these?

http://stackoverflow.com/questions/4165135/how-to-use-while-read-bash-to-read-the-last-line-in-a-file-if-there-s-no-new

brickstrap-unshare-harder.sh
dlech: I need to think about this part some more. Can you make this as a separate pull request?

Ubuntu has `unprivileged_userns_clone` enabled by default, so I think this is only going to be a problem on Debian (and non-Ubuntu relatives). And in Debian, I think it is more likely that root has a password and less likely that a user has sudo setup.

dlech:
This looks great! Thanks very much.

PR #22, opened 2016-01-05
Preliminary work on supporting a component based path mechanism.

cmacq2:
This work is the result of the discussion on issue #17:
https://github.com/ev3dev/brickstrap/issues/17

This commit provides preliminary support for:
 - The concept of a project (BR_PROJECT) (intended to be passed via commandline switch)
 - Validated root directory of the project (BR_PROJECT_DIR/ br_project_dir())
 - The concept of a list of 'selected components' (BR_COMPONENTS)
 - Validation of these components as directories within the project root
 - Resolving (map) base paths to find each applicable path in the project based on selected components. (br_list_*())
 - Callback interface for consuming these items (br_consume_path_list*())
 - Specifically: concatenating files with br_cat_files()
 - Specifically: iterating over files in all applicable base directories with br_consume_path_list_iterate_directories()

### DIFF
--- a/brickstrap-components.sh
+++ b/brickstrap-components.sh
@@ -1,0 +1,259 @@
+#!/bin/bash
+#
+# This file is part of brickstrap.
+#
+# brickstrap - create a foreign architecture rootfs using kernel namespaces,
+#              multistrap, and qemu usermode emulation and create a disk image
+#              using libguestfs
+#
+# Copyright (C) 2016 Johan Ouwerkerk <jm.ouwerkerk@gmail.com>
+#
+
+#
+# Note: this file is not meant to be executed, source it as a library of
+# functions instead. Variables used by the functions (other than stack) are
+# namespaced using the 'BR_' or 'BRP_' prefixes. Function names are namespaced
+# similarly, using the 'br_' and 'brp_' prefixes. See docs/namespacing.md for
+# more information on namespacing in brickstrap.
+#
+
+
+#
+# Validates the project name configured (via commandline arguments).
+# Additionally it sets another variable that contains the full path to the
+# project directory (to be able to distinguish between user input and what
+# brickstrap actually uses internally). See also: br_project_dir().
+#
+function brp_validate_project_name()
+{
+    if [ -z "$BR_PROJECT" ]; then
+        fail "No project specified (project name must not be empty)"
+    elif [ -r "$BR_PROJECT" -a -d "$BR_PROJECT" ]; then
+        BR_PROJECT_DIR=$(readlink -f "$BR_PROJECT")
+    elif [ -r "$(br_script_path)/$BR_PROJECT" ] && \
+        [ -d "$(br_script_path)/$BR_PROJECT" ]; then
+        BR_PROJECT_DIR=$(readlink -f "$(br_script_path)/$BR_PROJECT")
+    else
+        fail "Invalid project name (no such directory): '$BR_PROJECT'.
+Directory does not exist: '$BR_PROJECT'
+Directory does not exist: '$(br_script_path)/$BR_PROJECT'"
+    fi
+}
+
+#
+# Retrieve the absolute path to the project directory. This function maps
+# a validated project name to the full path of the project directory.
+# This function may be called once the project name has been validated; see
+# brp_validate_project_name().
+#
+function br_project_dir()
+{
+    [ -n "$BR_PROJECT_DIR" ] && echo -n "$BR_PROJECT_DIR"
+}
+
+#
+# Check that a given component name is valid. Valid component names correspond
+# to an existing subdirectory within the project directory.
+# This function requires that the project name has been successfully validated.
+# $1 the component name to check.
+#
+function brp_validate_component_name()
+{
+    if [ $# -eq 0 -o -z "$1" ]; then
+        fail "No component specified (component name must not be empty)"
+    elif [ ! -d "$(br_project_dir)/${1##/}" ]; then
+        fail "Not a valid component name: '$1'.
+Directory does not exist: '$(br_project_dir)/${1##/}'."
+    fi
+}
+
+#
+# Validate the configured component names (from commandline arguments).
+# This function requires that the project name has been successfully validated
+# first. See also: brp_validate_project_name().
+#
+function brp_validate_component_names()
+{
+    brp_iterate_components brp_validate_component_name || \
+        fail "At least one component is required"
+}
+
+#
+# Iterate over the configured component names (commandline arguments), invoking
+# callback for each component name. The calling convention is such that the
+# component name is passed as first argument to the callback, and any extra
+# arguments passed to this function are passed along to the callback aftet the
+# component name.
+#
+# $1: the callback to invoke.
+# $2...: optional: additional arguments passed to the callback following the
+#                  component name.
+#
+function brp_iterate_components()
+{
+    if [ -z "$BR_COMPONENTS" -o $# -eq 0 -o -z "$1" ]; then
+        return 1
+    else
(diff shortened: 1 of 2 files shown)


cmacq2:
This PR provides the 'guts' of work necessary for supporting the last proposal to come out of issue #17 

Code can be tested by faking the environment for ./tests/test-components.sh, e.g.:
```
# using a component of '.' provides backwards compatibility with projects that aren't updated yet.
BR_PROJECT=ev3-ev3dev-jessie BR_COMPONENTS=. ./tests/test-components.sh dir hooks
```

brickstrap-components.sh
```diff
@@ -0,0 +1,215 @@
...
+
+#
+# Note: this file is not meant to be executed, source it as a library of
+# functions instead. Variables used by the functions (other than stack) are
+# namespaced using the 'BR_' or 'BRP_' prefixes. Function names are namespaced
+# similarly, using the 'br_' and 'brp_' prefixes. See docs/namespacing.md for
+# more information on namespacing in brickstrap.
+#
+
+function brp_validate_project_name()
```
dlech: Can you add a brief description of this function? I had to stare at it a while before I realized what it was doing (which is nothing new from the current brickstrap).

brickstrap-components.sh
```diff
@@ -0,0 +1,215 @@
...
+{
+    [ -n "$BR_PROJECT" ] || fail "Project name must not be empty"
+    if [ -r "$BR_PROJECT" -a -d "$BR_PROJECT" ]; then
+        BR_PROJECT_DIR=$(readlink -f "$BR_PROJECT")
+    elif [ -r "$(br_script_path)/$BR_PROJECT" ] && \
+        [ -d "$(br_script_path)/$BR_PROJECT" ]; then
+        BR_PROJECT_DIR=$(readlink -f "$(br_script_path)/$BR_PROJECT")
+    else
+        fail "Invalid project name: '$BR_PROJECT'"
```
dlech: I would rather have the error message say something to the effect that it could not find a matching project source folder. If you want to get fancy, you could even check that the folder contains (some of) the required files.

brickstrap-components.sh
```diff
@@ -0,0 +1,215 @@
...
+
+function br_project_dir()
+{
+    [ -n "$BR_PROJECT_DIR" ] && echo -n "$BR_PROJECT_DIR"
+}
+
+function brp_validate_component_name()
+{
+    [ $# -eq 1 -a -n "$1" ] || fail "Component name must not be empty"
+    [ -d "$BR_PROJECT_DIR/$1" ] || fail "Not a valid component name: '$1'"
```
dlech: I would find it useful for the error message to say why it is not valid so that I know how to fix it.

dlech: something like "could not find component in $BR_PROJECT_DIR"

brickstrap-components.sh
```diff
@@ -0,0 +1,215 @@
...
+# (i.e. returned non-zero status code) or if another error occurs, the
+# br_consume_path_list*() function will return with a non-zero return code.
+#
+# This means something like this will print directories found using a given
+# 'printf' format:
+#
+# br_list_directories query | \
+#    br_consume_path_list printf 'found: %s\\n' || echo "error"
+#
+
```
dlech: This sounds a bit complicated. What do you plan on using it for?

dlech: I think it's mostly the word "consume" that is not making sense to me. What is being consumed?

brickstrap-components.sh
```diff
@@ -0,0 +1,215 @@
...
+#
+# Convenience function which concatenates files matching a given base path.
+#
+# $1 base path of the files to concatenate
+#
+function br_cat_files()
+{
+    br_list_paths "$1" -f | br_consume_path_list cat
+}
```
dlech: I like to always have newlines at the end of files.

brickstrap-components.sh
cmacq2: The purpose of the general mechanism is to execute a callback for each path found but be able to fail if any one of the callback invocations fail or if the mapping mechanism itself fails.

The thing being consumed is the produced list (line-by-line) of matching paths. If you like I could rename `consume` into `map`. Essentially the idea is: produce | map callback

For plain files you can see the `cat` version to cat them together (useful for e.g. `config` or `tar-exclude`). For directories you can see the `_iterate_directories` version (useful for e.g. running all hooks from all matching hooks directories, or examining all package files from all packages directories).

brickstrap-components.sh
cmacq2: This is basically copied from the existing validation logic for `$BOARDDIR` -- minus checking for a `multistrap conf`...

cmacq2:
Updated to integrate more comments/better error messages/and a newline.

With my explanation, do you still have objections to the 'consume' bits?

brickstrap-components.sh
dlech: This makes more sense now. Could we call the function `for_each` instead of `consume` or `map`? `consume` makes it seem to me like the list is being deleted as we go along (or we have a producer/consumer industrial data stream thing going on, which doesn't make sense in bash :smile:).

How is this better than `for f in $(br_list_directories query); do printf 'found: %s\\n' || echo "error" && break; done`? (Example might not be exactly correct, but you get the idea.)

cmacq2: > Could we call the function for_each instead of consume or map? consume makes it seem to me like the list is being deleted as we go along

Sure we could. I'd use `map` myself because it better fits the notion of taking a function and applying it to each item of a list a bit better (are you familiar with functional programming languages like Haskell?).

> How is this better than 

Well the example isn't quite correct as you already suspected: the error handling needs to be a bit more precise (to capture & return the specific error code if possible). From a DRY perspective that's not so good (remember that we'd have to have something like the loop at each call site, which is basically whenever brickstrap examines a file in the project directory). I figured, these are details we really don't want to be copy-pasted if we discover a bug or decide we want subtly different behaviour.

dlech: > are you familiar with functional programming languages like Haskell?

Sorry, but no. This is probably why I am having a hard time getting my head around this.

Let's call it ...for_each... and go with it. At the moment, I don't really care what the error code was, just that there was an error, but I suppose it may come in handy some day.

cmacq2: > Sorry, but no.
In case you're interested, it's even got its own Wikipedia article: https://en.wikipedia.org/wiki/Map_(higher-order_function)

In any case I've been revisiting the exact calling convention slightly because I overlooked the fact that when the eval/sourcing is on the RHS of the pipe symbol then you can't simply set variables the way you are used to. I think I've got a solution that gets rid of the pipe by using the `<<<` operator and adjusting the calling convention take a list as parameter to the `consume` function. 

I'll rename as per your request, the pattern will be `br_for_each_path*`

cmacq2: This is now fixed.

cmacq2:
I've pushed a new version of the PR which should fix the outstanding issues/objections.

That is: the `br_consume_path_list` functions have been renamed to `br_for_each_path`, and the calling convention has been changed to pass the path list as the first argument rather than via pipe. With this change things that are sourced through this mechanism (like config, hooks) can set variables that persist afterwards in the brickstrap.sh context, just like before.

My integration branch appears to be coming along nicely. I've taken the opportunity to drive-by fix a few niggles (did you know that the `-h` switch was actually broken?). When I'm happy about this/i.e. when I can bootstrap one of the example dirs again I'll polish it up a bit and post it as a follow up PR.

dlech:
Awesome! Merging.

> did you know that the -h switch was actually broken?

Thats kind of funny actually.

PR #23, opened 2016-01-11
Integrate component based path layering

cmacq2:
Integrate component based path layering into the main brickstrap script itself.

With this change brickstrap can take advantage of component base path scheme discussed in issue #17 and implemented in PR #22.
For details refer to:
Discussion: https://github.com/ev3dev/brickstrap/issues/17
Implementation PR: https://github.com/ev3dev/brickstrap/pull/22

Additional changes:

 - Namespace functions in brickstrap.sh
 - Move 'inline' code into proper standalone functions where possible
 - Fix -h switch
 - Enable other login shells besides bash

This change does **not** integrate the architecture/QEMU support from PR #21

### DIFF
--- a/brickstrap.sh
+++ b/brickstrap.sh
@@ -4,6 +4,7 @@
 #              multistrap, and qemu usermode emulation and create a disk image
 #              using libguestfs
 #
+# Copyright (C) 2016      Johan Ouwerkerk <jm.ouwerkerk@gmail.com>
 # Copyright (C) 2014-2015 David Lechner <david@lechnology.com>
 # Copyright (C) 2014-2015 Ralph Hempel <rhempel@hempeldesigngroup.com>
 #
@@ -19,39 +20,60 @@
 # http://twitter.com/kvz
 #
 
-### Configuration
+### Runtime
 #####################################################################
+set -e
+
+# Bash will remember & return the highest exitcode in a chain of pipes.
+set -o pipefail
 
 # Environment variables
 [ -z "${LOG_LEVEL}" ] && LOG_LEVEL="3" # 4 = debug -> 0 = fail
 
-# Commandline options. This defines the usage page
+#
+# Get the brickstrap source directory
+#
+function br_script_path()
+{
+    SCRIPT_PATH=$(readlink -f "$0")
+    SCRIPT_PATH=$(dirname "$SCRIPT_PATH")
+    echo -n "$SCRIPT_PATH"
+}
 
-read -r -d '' usage <<-'EOF'
-  usage: brickstrap [-b <board>] [-d <dir>] <command>
+. "$(br_script_path)/brickstrap-components.sh"
+
+# Commandline options. This defines the usage page
+BRP_USAGE=$(cat <<'EOF'
+Usage: brickstrap -p <project> [-c <component>] [-d <dir>] <command>
 
 Options
 -------
--b <board> Directory that contains board definition. (Default: ev3dev-jessie)
--d <dir>   The name of the directory for the rootfs. Note: This is also
-           used for the .tar and .img file names.
--f         Force overwriting existing files/directories.
--h         Help. (You are looking at it.)
+-c <component> Select components from the project directory.
+               Multiple components may be selected by specifying multiple '-c'
+               options; at least one component must be specified.
+-d <dir>       The name of the directory for the rootfs.
+               Note: This is also used for the .tar and .img file names.
+-p <project>   Directory which contains the brickstrap configuration (project).
+               This option is required and should occur exactly once.
+               Values are either a path to the project directory or the name of
+               an example project shipped with brickstrap by default.
+-f             Force overwriting existing files/directories.
+-h             Help. (You are looking at it.)
 
   Commands
   --------
   create-conf          generate the multistrap.conf file
-* simulate-multistrap  run multistrap with the --simulate option (for debugging)
+* simulate-multistrap  debug/dry-run of multistrap using its --simulate option
   run-multistrap       run multistrap (creates rootfs and downloads packages)
-  copy-root            copy files from board definition folder to the rootfs
+  copy-root            copy files from project definition folder to the rootfs
   configure-packages   configure the packages in the rootfs
-* run-hook <hook>      run a single hook in the board configuration folder
-  run-hooks            run all of the hooks in the board configuration folder
+* run-hook <hook>      run a single hook in the project configuration folder
+  run-hooks            run all of the hooks in the project configuration folder
 * create-rootfs        run all of the above commands (except *) in order
   create-tar           create a tar file from the rootfs folder
   create-image         create a disk image file from the tar file
-  create-report        run custom reporting script <board>/custom-report.sh
-* shell                run a bash shell in the rootfs
+  create-report        run custom reporting script <project>/custom-report.sh
+* shell [shell]        run the given shell in the rootfs (default is bash).
 * delete               deletes all of the files created by other commands
   all                  run all of the above commands (except *) in order
 
@@ -75,11 +97,12 @@ Options
   EV3DEV_RASPBIAN_MIRROR  Specifies the ev3dev/raspbian mirror used by apt
                           default: http://ev3dev.org/raspbian
                           (applies to create-conf only)
+
 EOF
+);
 
 ### Functions
 #####################################################################
-
 
 function _fmt () {
   color_info="\x1b[32m"
@@ -94,7 +117,7 @@ function _fmt () {
 
   color_reset="\x1b[0m"
   if [ "${TERM}" != "xterm" ] || [ -t 1 ]; then
-    # Don't use colors on pipes on non-recognized terminals
+    # Don't use colours when using pipes in unrecognised terminals
     color=""
     color_reset=""
   fi
@@ -107,9 +130,11 @@ function warn ()  { [ "${LOG_LEVEL}" -ge 2 ] && echo "$(_fmt warn) ${@}"  || tru
 function info ()  { [ "${LOG_LEVEL}" -ge 3 ] && echo "$(_fmt info) ${@}"  || true;         }
 function debug () { [ "${LOG_LEVEL}" -ge 4 ] && echo "$(_fmt debug) ${@}" || true;         }
 
-function help() {
-    echo >&2 "${@}"
-    echo >&2 "  ${usage}"
+function brp_help() {
+    if [ $# -ge 1 ]; then
+        echo >&2 "${@}"
+    fi
+    echo >&2 "${BRP_USAGE}"
     echo >&2 ""
 }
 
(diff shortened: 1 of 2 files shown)


brickstrap.sh
```diff
@@ -19,29 +20,51 @@
...
+Usage: brickstrap -p <project> [-c <component>] [-d <dir>] <command>
 
 Options
 -------
--b <board> Directory that contains board definition. (Default: ev3dev-jessie)
--d <dir>   The name of the directory for the rootfs. Note: This is also
-           used for the .tar and .img file names.
--f         Force overwriting existing files/directories.
--h         Help. (You are looking at it.)
+-b <board>     Directory that contains board definition. Alias for -p.
```
dlech: I'm not worried about backward compatibility. We can omit the -b option.

brickstrap.sh
```diff
@@ -19,29 +20,51 @@
...
+    echo -n "$SCRIPT_PATH"
+}
 
-read -r -d '' usage <<-'EOF'
-  usage: brickstrap [-b <board>] [-d <dir>] <command>
+. "$(br_script_path)/brickstrap-components.sh"
+
+# Commandline options. This defines the usage page
+BRP_USAGE=$(cat <<'EOF'
+Usage: brickstrap -p <project> [-c <component>] [-d <dir>] <command>
```
dlech: I know we discussed earlier about saving things like which qemu in the rootfs, but looking over the commands here, I am seeing a need to keep this info outside of the rootfs.

First, we need to expand the usage description to show which options are valid with which commands. Here is what I propose

    Usage: brickstrap <options> <command>
    Creation:
    brickstrap -p <project> -c <component> [ -c <component> ... ] [ -d <rootfs> ] { create-conf | create-rootfs | all }
    Incremental build:
    brickstrap -d <rootfs> { run-multistrap | copy-root | configure-packages | run-hooks | create-tar | create-image | create-report }
    Debug:
    brickstrap -d <rootfs> { simulate-multistrap | run-hook <hook> | shell [ <shell> ] }
    Destruction:
    brickstrap -d <rootfs> delete

brickstrap.sh
```diff
@@ -19,29 +20,51 @@
...
--b <board> Directory that contains board definition. (Default: ev3dev-jessie)
--d <dir>   The name of the directory for the rootfs. Note: This is also
-           used for the .tar and .img file names.
--f         Force overwriting existing files/directories.
--h         Help. (You are looking at it.)
+-b <board>     Directory that contains board definition. Alias for -p.
+-c <component> Select components from the board directory. Multiple components
+               may be selected by specifying multiple '-c' options, at least
+               one component is required.
+-d <dir>       The name of the directory for the rootfs.
```
dlech: Would it make more sense to call this `-r <rootfs>`?

brickstrap.sh
```diff
@@ -19,29 +20,51 @@
...
 
 Options
 -------
--b <board> Directory that contains board definition. (Default: ev3dev-jessie)
--d <dir>   The name of the directory for the rootfs. Note: This is also
-           used for the .tar and .img file names.
--f         Force overwriting existing files/directories.
--h         Help. (You are looking at it.)
+-b <board>     Directory that contains board definition. Alias for -p.
+-c <component> Select components from the board directory. Multiple components
```
dlech: s/board/project/

brickstrap.sh
```diff
@@ -19,29 +20,51 @@
...
+               Values are either a path to the project directory or the name of
+               an example project shipped with brickstrap by default.
+-f             Force overwriting existing files/directories.
+-h             Help. (You are looking at it.)
 
   Commands
   --------
   create-conf          generate the multistrap.conf file
-* simulate-multistrap  run multistrap with the --simulate option (for debugging)
+* simulate-multistrap  run multistrap with the --simulate option (to debug)
```
dlech: maybe better to get rid of `()` altogether and say `debug multistrap using 'multistrap --simulate'`

brickstrap.sh
```diff
@@ -51,7 +74,7 @@ Options
   create-tar           create a tar file from the rootfs folder
   create-image         create a disk image file from the tar file
   create-report        run custom reporting script <board>/custom-report.sh
-* shell                run a bash shell in the rootfs
+* shell <shell>        run the given shell in the rootfs (default is bash).
```
dlech: Use `[]` to show that `<shell>` is optional.

brickstrap.sh
dlech: The idea here is only have to specify the configuration options once using one of the "creation" commands. Since the most basic creation command is `create-conf` (which currently just creates `multistrap.conf`), we need to create a `brickstrap.conf` as well. It would just need to save the absolute path to the project and each component that was passed in. All other commands source brickstrap.conf (and fail if it does not exist).

brickstrap.sh
```diff
@@ -120,96 +146,172 @@ function help() {
...
+        while getopts "fhb:c:d:p:" BRP_OPT; do
+            case "$BRP_OPT" in
+            f) FORCE=true;;
+            h)
+                brp_help
+                exit 0
+            ;;
+            b|p)
+                if [ -z "$OPTARG" ]; then
+                    brp_help 'Empty project names are invalid'
```
dlech: I think it would be more natural to say "Must specify a project name."

brickstrap.sh
```diff
@@ -120,96 +146,172 @@ function help() {
...
+                    BR_PROJECT="$OPTARG"
+                else
+                    brp_help "Duplicate project: -$BRP_OPT '$OPTARG'.
+Project was: '$BR_PROJECT'"
+                    exit 1
+                fi
+            ;;
+            c)
+                if [ -z "$OPTARG" ]; then
+                    brp_help 'Empty component names are invalid'
```
dlech: Likewise, "Must specify a component name."

brickstrap.sh
```diff
@@ -120,96 +146,172 @@ function help() {
...
-        esac
-    done
-
-    shift $((OPTIND-1))
-
-    if [ "${cmd}" == "run-hook" ]; then
-      run_hook_arg=$1
-    fi
+            \:) # missing argument
+                brp_help "Missing argument for option: -$BRP_OPT"
```
dlech: OK, I see this should catch omitted arguments. An empty  argument would be "", so disregard my comments above.

brickstrap.sh
```diff
@@ -120,96 +146,172 @@ function help() {
...
+    debug "SCRIPT_PATH: $(br_script_path)"
 
-if [[ "$(sysctl -ne kernel.unprivileged_userns_clone)" == "0" ]]; then
-    fail "Unprivileged user namespace clone is disabled. Enable it by running" \
-        "'sudo sysctl -w kernel.unprivileged_userns_clone=1'."
-fi
+    export DEBIAN_FRONTEND=noninteractive
+    export DEBCONF_NONINTERACTIVE_SEEN=true
+    export LC_ALL=C LANGUAGE=C LANG=C
+    export PATH=$PATH:/usr/local/sbin:/usr/sbin:/sbin
```
dlech: I had not really thought about this before, be we might want to omit these when using `brickstrap shell`. We can save that for another day though.

brickstrap.sh
dlech: Actually, I take that back. It think we should leave it as `-d <dir>`. Currently, we create all files in the current directory and the actual rootfs has the same name as `<dir>`. However, I think it would be better if we contained all of the generated files in a parent directory. So, we have...


    <dir>/                        created by create-conf
      + rootfs/                   created by create-rootfs
      + brickstrap.conf           created by create-conf
      + multistrap.conf           created by create-conf
      + rootfs.tar                created by create-tar
      + <dir>.img                 created by create-image
      + <dir>-report.txt          created by create-report

brickstrap.sh
```diff
@@ -224,58 +326,28 @@ function create-conf() {
...
-                if [ -z "$line" ] || echo "$BLACKLIST_PACKAGES" | grep -q "$line"; then
-                    continue
-                else
-                    PACKAGES="${PACKAGES} $line"
-                fi
-            ;;
-            esac
-        done < "$f"
-    done
+    debug "BOARDDIR: $(br_project_dir)"
```
dlech: For debug output, I like the label to be whatever the variable is that we are printing, so in this case

    debug "br_project_dir: $(br_project_dir)"

brickstrap.sh
```diff
@@ -224,58 +326,28 @@ function create-conf() {
...
-        then
-               multistrapconf_cleanup=true
-        fi
-    done < $BOARDDIR/multistrap.conf
+    br_for_each_path "$(br_list_paths multistrap.conf -f)" \
+        brp_read_multistrap_conf_file
 }
 
-function simulate-multistrap() {
+function brp_simulate-multistrap() {
```
dlech: mixed underscore and dash is weird. can we just make it all `_`?

brickstrap.sh
```diff
@@ -349,70 +429,124 @@ function configure-packages () {
...
-        --exclude-from=${EXCLUDE_LIST} .
-    if [ -d "${BOARDDIR}/tar-only" ]; then
-      cp -r "${BOARDDIR}/tar-only/." "${ROOTDIR}/tar-only/"
-    fi
-    if [ -d "${ROOTDIR}/tar-only" ]; then
-      info "Adding tar-only files:"
-      ${CHROOTCMD} tar rvpf /host-rootfs/${TARBALL} -C /tar-only .
+        --exclude=host-rootfs --exclude=usr/bin/$(basename ${QEMU_STATIC}) \
+        --exclude=tar-only --exclude=${BRP_EXCLUDE_LIST##$ROOTDIR/} \
+        --exclude-from=${BRP_EXCLUDE_LIST##$ROOTDIR} .
```
dlech: This doesn't look right. Using the same parameter for both the `--exclude` and `--exclude-from` arguments?

cmacq2: The `--exclude` excludes a particular path, the `--exclude-from` reads `--exclude` values from a file. So we use both: one to get the configured `--exclude-from` filter and the other to exclude the filter. This behaviour is not new either. :)

brickstrap.sh
cmacq2: Sure. Will do.

brickstrap.sh
cmacq2: I'd rather remove those old debug statements, I had simply updated them to the 'correct' value.

brickstrap.sh
cmacq2: Yes. Also, yes: let's save that for a follow up PR because the changeset is big enough as it its. ;)

brickstrap.sh
dlech: Sanity checks are as follows...

* Creation commands:
    * Checks for valid project and component source directories
    * Checks that `<dir>` does not exist
       * If no `-f` option, fails with error.
       * if yes `-f` option, runs `delete` command before proceeding
* Incremental build
    * Checks that `<dir>` is sane, e.g. has `brickstrap.conf`
    * Checks that project and components in brickstrap.conf are still valid
    * Checks that output of command does not already exist.
       * if no `-f` option, fail with error
      * if yes `-f` option, delete previous output and then proceed
* Debug:
    * check that `<dir>` is sane. e.g. has multistrap.conf for simulate-multistrap or has rootfs for run-hook/shell
* Delete:
    * check that `<dir>` exists

brickstrap.sh
cmacq2: Actually, no it doesn't: turns out this is dead code. I added this based on the getopt C API which uses this convention, but turns out that bash getopt simply prints:
> ../brickstrap/brickstrap.sh: option requires an argument -- p
Invalid option specified

Followed by the usage

Meanwhile an illegal option like `-Z` gets you:

> ../brickstrap/brickstrap.sh: illegal option -- Z
Invalid option specified

That means "Invalid option specified" is also null-content. I'll update the PR to correct this and remove dead-code/null-content.

Note that it is explicitly possible to pass empty things using quotes `brickstrap.sh -p ""`. That's what the checks for empty names are about.

brickstrap.sh
cmacq2: Sure. Will remove it.

brickstrap.sh
cmacq2: Let's leave these kind of behaviour changes out of the scope of this PR. There's a lot of scope for bugs and unintended consequences:

 * what are the semantics of the -f flag? 
 * what if you do want a new config? 
 * how do you know all settings are preserved and not just the ones which are hit by the code path of the current command? 
 * how do you avoid the 'history' simply growing and growing endlessly, leading to worse and worse start up time and needless waste of disk space?

Perhaps you could setup an issue for exploring this/tracking these changes?

brickstrap.sh
dlech: > This behaviour is not new either.

I should know better, I'm the one that put it there. Thanks for reminding me how it works.



> What would be a useful for a follow-up PR is to nest things like tar-only, tar-exclude and host-rootfs underneath a brickstrap subdirectory in $ROOTDIR.

:+1:

brickstrap.sh
dlech: Let's leave them for now. If I find that I don't need them when transitioning the multiple ev3dev projects to a single project with components, then we can remove them.

brickstrap.sh
cmacq2: I agree, this would also make brickstrap behave better when invoked from arbitrary directories.

brickstrap.sh
cmacq2: Noted, will do.

brickstrap.sh
cmacq2: Agreed.

brickstrap.sh
cmacq2: Will do.

brickstrap.sh
dlech: #24

brickstrap.sh
dlech: #25

brickstrap.sh
dlech: Based on your comment below, you are going to save this for another PR?

cmacq2: Yes. Let's avoid scope creep in this PR. :)

cmacq2:
Updated to integrate feedback. I've tested this with a fresh run of a project build (using 'all') to check that the function renaming didn't accidentally break things.

brickstrap.sh
dlech: #26

cmacq2:
Do you have any further comments? Or can the PR be merged in as it stands now?

dlech:
I would like to actually test it myself first. Haven't got around to that yet.

cmacq2:
Okay. Tip: pass `-c .` if you haven't converted your project directories yet.

cmacq2:
I've tightened validation of project & component names.
This change disallows the following 'project' paths:

 - brickstrap source directory
 - tests/ subdirectory of brickstrap
 - docs/ subdirectory of brickstrap

Additionally trivially duplicate component names are now ignored, and component names which attempt to 'escape' the project directory like `../` are disallowed.

There's still the edge case of someone specifying a non-trivial duplicate component name, for example: `-c name1/../name2 -c name2`

brickstrap.sh
```diff
@@ -311,32 +394,26 @@ function configure-packages () {
...
-                    blacklisted="true"
-                    info "skipping ${script##$script_dir} (blacklisted)"
-                    break
-                fi
-            done < "${BOARDDIR}/preinst.blacklist"
+    BRP_script_dir="${ROOTDIR}/var/lib/dpkg/info"
+    BRP_preinst_blacklist="$(br_cat_files preinst.blacklist)"
+    for BRP_script in ${BRP_script_dir}/*.preinst; do
+        if echo "$BRP_preinst_blacklist" | \
+            grep -q "$(basename "$BRP_script" .preinst)"; then
```
dlech: The name matching needs to be fixed here. When testing with the `ev3-ev3dev-jessie` project, `systemd.preinst` was incorrectly identified as blacklisted. This is probably triggered by the fact that `ev3-systemd` is in the preinst.blacklist.

cmacq2: Good call. Using `egrep -q "^$(basename "$BRP_script" .preinst)\$"` should do it. Will amend the PR.

cmacq2:
Bug fix: match blacklisted scripts against whole lines, so substring matches should no longer lead to false positives.

brickstrap.sh
dlech: Will this handle package names with a `.` in them?

cmacq2: I don't see why it shouldn't? `basename .preinst` basically gives you the `basename` minus `.preinst` suffix. That should be equivalent to `$line` from the before-diff situation, and that is what we feed into the pattern to `egrep`. So the `egrep` query should amount to the equivalent of `^${line}\$`, e.g.: `$line=foo.bar.baz` expands to `^foo.bar.baz$`

dlech: Right, but `.` is a special regex char, so `foo1.0` would match `foo100` too. The odds are low that there are actually package names where this would happen are probably pretty low, but it *could* happen, especially with kernel packages that have long version strings in the package name.

cmacq2: Yes I realised. But it turns out grep supports anchors anyway, so just use that. I updated the PR accordingly, again.

dlech: and `.` is still a regex character in regular grep, so we can still get bad matches

dlech: How about `grep -F`?

dlech: Make that `grep -xF`

cmacq2: You are quite right, of course. 

So I've broken out ~~the sledgehammer~~ use this construct: `grep -Pq "^\Q<pattern>\E\$"`

`-P` enables Perls style regexes, and the `\Q` and `\E` things are for quoting a literal inside a regex, regardless of how weird or wonderful or special the characters might be. ...

cmacq2: Your solution is better, cleaner. I'll update yet again: fgrep -x

cmacq2:
Updated to use the `fgrep -xq "<pattern>"` construct. This is cleaner than the previous 'fixes'..

dlech:
Shouldn't we be using `fgrep` everywhere else we are grepping a variable string?

cmacq2:
Unless we explicitly use the anchors... `fgrep`/`grep -F` doesn't do those. I'll prepare a new update. I checked: it's just brickstrap.sh that needs to be adjusted.

dlech:
Thanks for fixing up the greps. I'm running it one more time to make sure that changing the greps didn't break anything. In the meantime, the readme needs to be updated for the command line switch changes.

cmacq2:
I'd like to postpone the readme until we've also integrated the architecture/QEMU related changes into brickstrap.sh.

dlech:
Thanks!

Issue #18, opened 2015-12-28
Let's not hard-depend on QEMU for ARM, enable other arches and native builds as well


cmacq2:
Splitting from discussion in issue #17
 
> Regarding non-32-bit-arm architectures, I think we just use the existing ARCH variable from the config files to pick the correct qem-user-static (or none at all). But you should open a separate issue for that.

Yes. I would go with the following refinement of this basic idea, though:

 1. Permit the user to set the correct QEMU directly in their config file.
 2. If the config file doesn't specify ARCH set ARCH to the value learned from -A if it is available.
 3. If ARCH is still empty, assume a native build. Possibly recognise the special value "native" to assume a native build also.
 4. If the user hasn't specified the right QEMU manually, use the value of ARCH to pick a suitable default.
 5. If it's a native build set ARCH to the host architecture.

This way the user can also use QEMU which they built themselves, or if no official Debian is available for the arch yet. Useful when chasing QEMU bugs like missing syscalls, or if you are trying to port Debian to something new/obscure ...

cmacq2:
Time to close this issue: the necessary support things have landed and been integrated now.

Issue #24, opened 2016-01-11
Use brickstrap.conf to persist options between invocations of brickstrap.


dlech:
This conversation started in PR #23, so here's a copy...

@dlech:

> I know we discussed earlier about saving things like which qemu in the rootfs, but looking over the commands here, I am seeing a need to keep this info outside of the rootfs.
>
> First, we need to expand the usage description to show which options are valid with which commands. Here is what I propose
>
>     Usage: brickstrap <options> <command>
>     Creation:
>     brickstrap -p <project> -c <component> [ -c <component> ... ] [ -d <dir> ] [ -f ] { create-conf | create-rootfs | all }
>     Incremental build:
>     brickstrap -d <dir> [ -f ] { run-multistrap | copy-root | configure-packages | run-hooks | create-tar | create-image | create-report }
>     Debug:
>     brickstrap -d <dir> { simulate-multistrap | run-hook <hook> | shell [ <shell> ] }
>     Destruction:
>     brickstrap -d <dir> delete

> The idea here is only have to specify the configuration options once using one of the "creation" commands. Since the most basic creation command is create-conf (which currently just creates multistrap.conf), we need to create a brickstrap.conf as well. It would just need to save the absolute path to the project and each component that was passed in. All other commands source brickstrap.conf (and fail if it does not exist).

> Sanity checks are as follows...
> 
> * Creation commands:
>     * Checks for valid project and component source directories
>     * Checks that `<dir>` does not exist
>        * If no `-f` option, fails with error.
>        * if yes `-f` option, runs `delete` command before proceeding
> * Incremental build
>     * Checks that `<dir>` is sane, e.g. has `brickstrap.conf`
>     * Checks that project and components in brickstrap.conf are still valid
>     * Checks that output of command does not already exist.
>        * if no `-f` option, fail with error
>       * if yes `-f` option, delete previous output and then proceed
> * Debug:
>     * check that `<dir>` is sane. e.g. has multistrap.conf for simulate-multistrap or has rootfs for run-hook/shell
> * Delete:
>     * check that `<dir>` exists

dlech:
@cmacq2:

> Let's leave these kind of behaviour changes out of the scope of this PR. There's a lot of scope for bugs and unintended consequences:
> 
>  * what are the semantics of the -f flag? 
>  * what if you do want a new config? 
>  * how do you know all settings are preserved and not just the ones which are hit by the code path of the current command? 
[...]

PR #27, opened 2016-01-13
Integrate support for arbitrary architectures in brickstrap itself.

cmacq2:
With this change brickstrap can take advantage of the work done to support other architectures besides 'arm'.

For reference see:
Discussion: issue #18
https://github.com/ev3dev/brickstrap/issues/18
Implementation: PR #21 
https://github.com/ev3dev/brickstrap/pull/21

Additional changes:
  - Fix display of exclude list in output to include a newline (without adding it to the actual exclude list)

Note: this change does **not** address updating the `ev3dev` board directories/example projects.

### DIFF
--- a/brickstrap-utils.sh
+++ b/brickstrap-utils.sh
@@ -57,7 +57,7 @@ function brp_is_native()
     [ $# -eq 1 -a -n "$1" ] && if [ "$(dpkg --print-architecture)" = "$1" ]; \
     then
         return 0
-    elif [ "$1" = "native" ]; then
+    elif [ "$1" = "none" ]; then
         return 0
     else
         return 1
@@ -119,7 +119,7 @@ function brp_guess_qemu()
     [ $# -eq 1 -a -n "$1" ] && \
     if brp_is_native "$1"; then
         return 0
-    elif
+    else
         which "qemu-$(brp_resolve_qemu_alias "$1")-static"
     fi
 }
@@ -130,32 +130,45 @@ function brp_guess_qemu()
 #  - prefer an explicitly configured QEMU (QEMU_STATIC or BR_QEMU)
 #  - next, attempt to auto-detect QEMU through binfmt support scripts/tools
 #  - finally, guess QEMU based on contents of ARCH, if available
-# Returns 1 if QEMU was explicitly configured by the user.
-# Fails if no QEMU can be determined.
+# Returns 0 if no QEMU is required.
+# Returns 254 if QEMU was explicitly configured by the user.
+# Returns 255 if QEMU was auto-detected.
+# BRP_BINFMT is set to the QEMU which is found or to the empty string if no
+# emulator is required. Fails if an emulator cannot be determined.
 #
-function brp_determine_qemu()
+function brp_determine_qemu_impl()
 {
-    BRP_BINFMT=$(brp_validate_qemu)
-
-    # was BRP_BINFMT configured via environment or commandline option argument?
-    if [ $? -eq 0 ]; then
+    BRP_BINFMT=""
+    # is BRP_BINFMT configured via environment or commandline option argument?
+    if brp_validate_qemu; then
         if [ -n "$BRP_BINFMT" ]; then
-            echo -n "$BRP_BINFMT"
-            return 1
-        else # explicit 'native' configuration.
-            return 0
+            info "Using configured QEMU: '$BRP_BINFMT'"
+            return 254
+        else
+            info "Not using QEMU: configuration indicates it is not required"
         fi
+    # is BRP_BINFMT explicitly not configured yet.
+    elif brp_auto_detect_qemu; then
+        if [ -n "$BRP_BINFMT" ]; then
+            info "Using auto-detected QEMU: '$BRP_BINFMT'"
+            return 255
+        else
+            info "Not using QEMU: detected or guessed that it is not required"
+        fi
+    else
+        fail "Unable to determine QEMU!"
     fi
+}
 
-    BR_BINFMT=$(brp_find_binfmt)
+function brp_auto_detect_qemu()
+{
+    BRP_BINFMT=$(brp_find_binfmt)
     if [ $? -eq 0 ]; then
         # double check that the binfmt thingy refers to what appears to be QEMU
         # it could be something silly like a misconfigured binfmt support.
-        [ -z "$BR_BINFMT" ] || brp_check_binfmt_qemu "$BR_BINFMT" || \
-            warn "This may not be a QEMU binary after all: '$BR_BINFMT'.
+        [ -z "$BRP_BINFMT" ] || brp_check_binfmt_is_qemu "$BRP_BINFMT" || \
+            warn "This may not be a QEMU binary after all: '$BRP_BINFMT'.
 Check your binfmt configuration and/or configure QEMU to silence this warning."
-        echo -n "$BR_BINFMT"
-        return 0
     else
         # canonical 'ARCH' var is set
         # Fall back to guessing games involving $ARCH
@@ -167,7 +180,7 @@ Check your binfmt configuration and/or configure QEMU to silence this warning."
         if [ -n "$ARCH" ]; then
             warn "Unable to auto-detect required QEMU based on binfmt support.
 Reverting to guessing QEMU; install binfmt-support packages to avoid it."
-            brp_guess_qemu "$ARCH"
+            BRP_BINFMT=$(brp_guess_qemu "$ARCH")
         # arch not set, report failure of binfmt support mechanism as fatal
         elif [ ! -x /usr/sbin/update-binfmts ]; then
             fail "/usr/sbin/update-binfmts is not executable.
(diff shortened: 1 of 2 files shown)


cmacq2:
Tested this with an example project of mine. Note that in order to test this change you need to update your project as follows:

 * Convert each call to `${CHROOTCMD}` to use `br_chroot` instead
 * Likewise: convert each `${CHROOTBINDCMD}` to use `br_chroot_bind` instead
 * Unlikely but possible: convert each `${USER_UNSHARE}` to use `br_unshare` instead
 * Convert each instance of `$BOARDDIR` to use `$(br_project_dir)` instead.

This probably amounts to updating your `config`, `multistrap.conf` and hooks.

brickstrap.sh
```diff
@@ -57,6 +58,18 @@ Options
...
                an example project shipped with brickstrap by default.
+-Q <qemu>      Specify which QEMU binary to use for emulation of foreign
+               instruction sets; this may be used to override auto-detected
+               QEMU.
+
+               The <qemu> value must be either the path to a binary or
+               the name of a Debian or QEMU architecture or the special string
+               'native'. If <qemu> corresponds to a binary it is used
+               unconditionally, without further validation. If <qemu> is an
+               architecture which matches the host architecture or 'native'
```
dlech: Would this make more sense to say `none`here rather than `native` to get the idea across that qemu is not used in this case?

brickstrap.sh
```diff
@@ -57,6 +58,18 @@ Options
                This option is required and should occur exactly once.
                Values are either a path to the project directory or the name of
                an example project shipped with brickstrap by default.
+-Q <qemu>      Specify which QEMU binary to use for emulation of foreign
+               instruction sets; this may be used to override auto-detected
+               QEMU.
```
dlech: Could we just say "Optionally override the qemu binary used to emulate a foreign architecture." I think this makes it a bit more clear that this is an optional parameter and is bit more succinct.

brickstrap.sh
cmacq2: Sure.

brickstrap.sh
cmacq2: Will do.

cmacq2:
How about this new version?

brickstrap-utils.sh
```diff
@@ -183,27 +196,25 @@ Do you have necessary binfmt packages installed?"
...
+# The variable 'BRP_BINFMT' is set to the passed QEMU interpreter if valid.
+# Returns 255 if no QEMU was configured. Fails if the QEMU settings are
+# misconfigured.
 #
 function brp_validate_qemu()
 {
     # check environment variable first, may have been set by config
     if [ -n "$QEMU_STATIC" -a -x "$QEMU_STATIC" ]; then
-        echo -n "$(readlink -f $QEMU_STATIC)"
+        BRP_BINFMT="$(readlink -f $QEMU_STATIC)"
```
dlech: Can we just drop the un-namespaced `QEMU_STATIC` variable?

brickstrap.sh
```diff
@@ -352,11 +373,11 @@ function brp_run_multistrap() {
...
     if [ ! -z ${FORCE} ] && [ -d "${ROOTDIR}" ]; then
         warn "Removing existing rootfs ${ROOTDIR}"
-        ${USER_UNSHARE} -- rm -rf ${ROOTDIR}
+        brp_unshare -- rm -rf ${ROOTDIR}
     fi
-    ${USER_UNSHARE} -- multistrap ${MSTRAP_SIM} --file "${MULTISTRAPCONF}" \
+    brp_unshare -- multistrap ${MSTRAP_SIM} --file "${MULTISTRAPCONF}" \
         --dir ${ROOTDIR} --no-auth
-    cp ${QEMU_STATIC} ${ROOTDIR}/usr/bin/
+    brp_setup_qemu_in_rootfs
```
dlech: You didn't update the complementary statement in line 546 that excludes this file from the tar. Line 546 is still using the `QEMU_STATIC` variable.

brickstrap.sh
```diff
@@ -516,12 +537,12 @@ function brp_create_tar() {
...
     BRP_EXCLUDE_LIST="${ROOTDIR}/tar-exclude"
-    br_cat_files tar-exclude | tee "$BRP_EXCLUDE_LIST"
+    br_cat_files tar-exclude | tee "$BRP_EXCLUDE_LIST" && echo "" # add newline
 
     # need to generate tar inside fakechroot
     # so that absolute symlinks are correct
 
-    ${CHROOTCMD} tar cpf /host-rootfs/${TARBALL} \
+    br_chroot tar cpf /host-rootfs/${TARBALL} \
         --exclude=host-rootfs --exclude=usr/bin/$(basename ${QEMU_STATIC}) \
```
dlech: Ah, here is the line 546 I was talking about.

brickstrap.sh
cmacq2: Which should be OK, because `$QEMU_STATIC` is updated by `brp_setup_qemu_in_rootfs` during the copying.

dlech: Unless you run `create-tar` separately. But I guess we'll save that for #24.

brickstrap-utils.sh
cmacq2: Currently `QEMU_STATIC` may be set by the `config` or environment and takes precedence over `BR_QEMU` for that reason. It may be wortwhile to actually increase support for this to include the architecture->binary mapping for this variable as well in `brp_validate_qemu`.

Also it is somewhat useful information to have for `custom-report.sh`.

brickstrap.sh
```diff
@@ -214,9 +238,10 @@ function brp_sanity_check_cli_options()
 
 function brp_validate_cli_options()
 {
-    brp_sanity_check_cli_options "$@" && \
-    brp_validate_project_name && \
+    brp_sanity_check_cli_options "$@"
+    brp_validate_project_name
     brp_validate_component_names
+    brp_validate_qemu || [ $? -eq 255 ] # no QEMU specified = 255
```
dlech: It seems that `brp_validate_qemu` will allow anything other than an empty string when specifying the `-Q` option. Is that the intended behavior? It seems like it should check that an actual matching binary could be found at this point. Otherwise, you get all the way through downloading and unpacking packages before brickstrap fails, which wastes lots of time.

brickstrap.sh
cmacq2: You are right, and right again: let's save that for #24. Currently a somewhat awkward workaround exists: setting `QEMU_STATIC` manually before running `create-tar`.

brickstrap-utils.sh
dlech: So `QEMU_STATIC` is user-specified and `BR_QEMU` is strictly internal? I'm having a hard time following things with 3 variable that seem to essentially hold the same information. These two and `BRP_BINFMT`.

cmacq2: The distinctions are as follows:
 * `BR_QEMU` is namespaced as brickstrap+public, and it holds the raw `-Q` value. Project configurations may inspect this value but may not alter it. Because of the namespacing it's also not really valid to set this in the environment prior to invoking brickstrap.
 * `BRP_BINFMT` corresponds to the eventually determined binfmt binary in the *host* file system. This is not necessarily the same when viewed from inside the chroot. Examples of where this non-obvious detail matters: developer QEMU builds that live in home directories or `/usr/local/bin`. Hence, the copy logic contains comments to that effect. `BRP_BINFMT` is namespaced as brickstrap+private, so project configuration should not examine its value and in fact it is meant to be local to the `-utils.sh` module. I'll see about introducing a `brp_` namespaced getter, as a side effect of that it should be possible to also fix `create-tar` properly, again.
 * `QEMU_STATIC` is a globally namespaced variable which the project configuration may inspect *and* alter as it sees fit. Eventually it is set to the QEMU binary in the `rootfs`, as seen from inside the chroot.

dlech: OK, I found a comment that says `BR_QEMU` is the value passed from the `-Q` command line option. If this is correct, can we rename it to `BR_OPTION_QEMU`? (and in the future follow the `BR_OPTION_XYZ` pattern for all options).

brickstrap.sh
dlech: I think what I would like to have happen here is that if `$QEMU_STATIC` or `-Q something` is specified, resolve it and make sure that a source qemu file actually exists. If qemu is not specified in any way, make sure `update-binfmts` exists by running `update-binfmt --version`. This way we can fail as early as possible in most cases.

brickstrap-utils.sh
dlech: Furthermore, could we call `BRP_BINFMT` something like `BRP_SOURCE_QEMU` or `BRP_QEMU_SOURCE_PATH`?

brickstrap.sh
cmacq2: No can do. For `update-binfmts` to work you have to have a target binary. For you to have a target binary requires you to have performed the multistrap first.

The option assignment was broken, is now fixed, so validation works now. If you explicitly pass a `-Q` option or set `QEMU_STATIC` a misconfiguration is caught early (as intended).

dlech: I know that you can't run `update-binfmts --find`, but you can run `update-binfmts --version` to make sure that `update-binfmts` itself exists.

cmacq2: We test that using the `-x` option currently and anyway it is valid for it not to exist even if no QEMU is specified: native builds...

cmacq2:
There was a bug in option parsing causing validation to be skipped and also on reflection QEMU_STATIC should probably support the architecture mapping mechanism as well.

... I've amended the PR accordingly.

I've also added appropriate 'getter' functions, and use those inside `create-tar` as well.

brickstrap.sh
dlech: OK. I missed the `-x`.

dlech:
The command line parsing seems to be working as expected now. I'm going to run through a full build to see what happens.

cmacq2:
A note on `BR_QEMU`, actually I would prefer not to rename it to remain consistent with `BR_PROJECT` and `BR_COMPONENTS` which are similar (the raw `-p` and `-c` option values).

It's probably easier to remember this way:
 
 - getter functions to get the sane/sanitised paths (br_project_dir/br_get_rootfs_qemu/br_get_host_qemu) 
 - `BR_` variables to hold raw values.

brickstrap.sh
cmacq2: That's at line 86 of `brickstrap-utils.sh` in: `brp_find_binfmt`.

It's only checked once we actually can do the detection of the binfmt, but that makes little difference because as I said it's valid for it not be there (we can recover through guessing from `ARCH` if it is sane).

dlech:
Alright. That's fine as long as we start a doc that explains what the common variables and functions are that one might want to use in their hooks.

cmacq2:
Which should probably be done concurrently with the necessary readme updates...
... By the way, did your test succeed?

dlech:
The build looked good (other than I forgot to update the hooks like you told me to the first time around). I've pushed this along with my changes.

Issue #26, opened 2016-01-11
Create parent directory to contain all generated brickstrap files


dlech:
Currently, brickstrap creates files and a rootfs directory in the current directory. I think it would be better if we contained all of the generated files in a parent directory. So, we have...

    <dir>/                        created by create-conf
      + rootfs/                   created by create-rootfs
      + brickstrap.conf           created by create-conf
      + multistrap.conf           created by create-conf
      + rootfs.tar                created by create-tar
      + <dir>.img                 created by create-image
      + <dir>-report.txt          created by create-report

dlech:
Fixed in #28

PR #28, opened 2016-01-15
Extract common code, and cleanup various directory paths.

cmacq2:
This change introduces the new brickstrap-argv.sh module.

Changes:

 - Change the -d option to be 'destdir' as opposed to 'rootfs' directory.
   All output of brickstrap should now be generated as files or subdirectorys located within the given 'destdir'.
 - Extract common option parsing code into utility functions
 - Re-organise brickstrap artefacts inside rootfs/chroot.
   Artefacts like tar-only directory or tar-exclude file are now confined to a 'brickstrap' directory hierarchy.
   This way there is a lot less arbitrary poking inside the rootfs going on.
 - Introduce 'images' directory for storing rootfs image files.
   This paves the way for generating multiple images with arbitrary partition layouts.
 - Introduce -I option to specify image base name.
   By default the basename of 'destdir' is used with an appended timestamp.
 - Introduce getter function to lookup various paths, and base paths.
 - Remove numerous globally namespaced variables, notably ROOTDIR.
   (There are 'getter' functions which provide same information more safely now).
 - Introduce -P option for blacklisting individual packages.
   This is a more convenient alternative to BLACKLIST_PACKAGES
 - Introduce -B option for blacklisting entire package files.
   This is a more convenient alternative to BLACKLIST_PACKAGE_FILES
 - Update brickstrap-utils.sh module to take advantage of the changes.
 - Update brickstrap.sh to take advantage the changes.
 - Fix image generation logic to not require a temporary file + mv step, to avoid name clashes.

This change resolves issue #26 (through destdir):
https://github.com/ev3dev/brickstrap/issues/26

See also:
https://github.com/ev3dev/brickstrap/pull/23#discussion_r49371653

### DIFF
--- a/brickstrap-argv.sh
+++ b/brickstrap-argv.sh
@@ -1,0 +1,249 @@
+#!/bin/bash
+#
+# This file is part of brickstrap.
+#
+# brickstrap - create a foreign architecture rootfs using kernel namespaces,
+#              multistrap, and qemu usermode emulation and create a disk image
+#              using libguestfs
+#
+# Copyright (C) 2016 Johan Ouwerkerk <jm.ouwerkerk@gmail.com>
+#
+
+#
+# Note: this file is not meant to be executed, source it as a library of
+# functions instead. Variables used by the functions (other than stack) are
+# namespaced using the 'BR_' or 'BRP_' prefixes. Function names are namespaced
+# similarly, using the 'br_' and 'brp_' prefixes. See docs/namespacing.md for
+# more information on namespacing in brickstrap.
+#
+
+#
+# Set a single-valued option, if it hasn't been set already.
+# $1: name of the variable to set
+# $2: current value of the same.
+# $3: new value to set
+# $4: failure 'mode' string, values are interpreted as follows:
+#       * Starting with '-' implies a commandline option: failure is
+#         handled using a combination of 'brp_help' and 'exit 1'
+#       * A value of 'error', 'warn', 'info', and 'debug': pass a message
+#         to the correspondig logging function and continue silently.
+#       * A value of 'fail': log a message and terminates the script.
+#       * Any other value implies handling as per default: a message is logged
+#         using 'fail' (and the script is terminated), unless the error can be
+#         recovered from tivrially by ignoring the new value to set. In that
+#         case, a warning message is logged instead and the script continues
+#         silently.
+# $5: short descriptive 'name' for the setting being altered.
+#
+function brp_set_single_value_opt()
+{
+    if [ -z "$3" ]; then
+        case "$4" in
+        -*)
+            brp_help "Empty values are invalid as '$5' setting"
+            exit 1
+        ;;
+        warn|info|error|debug|fail)
+            $4 "Empty values are invalid as '$5' setting"
+        ;;
+        *)  fail "Empty value are invalid as '$5' setting";;
+        esac
+    elif [ -z "$2" ]; then
+        eval "$1=\"$3\""
+    elif [ "$2" = "$3" ]; then
+        case "$4" in
+        -*) warn "Ignoring duplicate value for '$5' setting: $4 '$3'"
+        ;;
+        warn|info|error|debug|fail)
+            $4 "Ignoring duplicate value for '$5' setting: '$3'"
+        ;;
+        *)  warn "Ignoring duplicate value for '$5' setting: '$3'"
+        ;;
+        esac
+    else
+        case "$4" in
+        -*)
+            brp_help "Duplicate value for '$5' setting: $4 '$3'.
+Previous setting: '$2'"
+            exit 1
+        ;;
+        warn|info|error|debug|fail)
+            $4 "Duplicate value for '$5' setting: '$3'.
+Previous setting: '$2'"
+        ;;
+        *)  fail "Duplicate value for '$5' setting: '$3'.
+Previous setting: '$2'";;
+        esac
+    fi
+}
+
+#
+# Append to a multi-valued option (list).
+# $1: name of the variable to append to
+# $2: current value of the same.
+# $3: new value to add
+# $4: failure 'mode' string, values are interpreted as follows:
+#       * Starting with '-' implies a commandline option: failure is
+#         handled using a combination of 'brp_help' and 'exit 1'
+#       * A value of 'error', 'warn', 'info', and 'debug': pass a message
+#         to the correspondig logging function and continue silently.
+#       * A value of 'fail': log a message and terminates the script.
+#       * Any other value implies handling as per default: a message is logged
+#         using 'fail' (and the script is terminated), unless the error can be
+#         recovered from tivrially by ignoring the new value to set. In that
+#         case, a warning message is logged instead and the script continues
+#         silently.
+# $5: short descriptive 'name' for the setting being altered.
+#
+function brp_set_multi_value_opt()
+{
+    if [ -z "$3" ]; then
+        case "$4" in
+        -*)
+            brp_help "Empty values are invalid as '$5' setting"
+            exit 1
+        ;;
+        warn|info|error|debug|fail)
+            $4 "Empty values are invalid as '$5' setting"
+        ;;
+        *)  fail "Empty value are invalid as '$5' setting";;
+        esac
+    elif [ -z "$2" ]; then
+        eval "$1=\"'$3'\""
+    elif echo "$2" | fgrep -q "'$3'"; then
+        case "$4" in
+        -*)
+            warn "Ignoring duplicate value for '$5' setting: $4 '$3'"
+        ;;
+        warn|info|error|debug|fail)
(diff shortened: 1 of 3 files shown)


cmacq2:
Note that this change removes a number of variables from the global namespace, notably `ROOTDIR`. Corresponding updates to the example projects are **not** part of this PR and must be incorporated separately:
 * ROOTDIR -> $(br_rootfs_dir)
 * IMAGE -> $BR_IMAGE_BASE_NAME (probably what you want)
 * Any others? See the 'getter' functions down in `brickstrap-argv.sh`.

Likely ROOTDIR is the big one, use:
```
grep -R ROOTDIR <project>
```
To find out which files need updating.

cmacq2:
Any comments?

dlech:
Sorry, I've been deep into other things. I'll try to have a look soon.

dlech:
> Introduce 'images' directory for storing rootfs image files. This paves the way for generating multiple images with arbitrary partition layouts.

What is the use case for multiple images from a single rootfs? It seems to me that if you want a different image, you would have to build a new rootfs. I would just as soon have the (single) image not in a directory by itself.

dlech:
> Introduce -P option for blacklisting individual packages. This is a more convenient alternative to BLACKLIST_PACKAGES

If we are going to get into this sort of fine-grained options, I would like to have long option names. Of course, that is out of the scope of this pull request.

Also, I would like to choose more obscure letters for the short form of blacklisting so that we can save the more common letters (P and B) for future options where they might make more sense.

brickstrap-argv.sh
```diff
@@ -0,0 +1,237 @@
+#!/bin/bash
+#
+# This file is part of brickstrap.
```
dlech: Just throwing this out there... Should we be writing brickstrap in python or some other language that has a decent arg parsing library so that we don't have to do all of this ourselves?

brickstrap.sh
```diff
@@ -40,35 +40,43 @@ function br_script_path()
...
+Usage: brickstrap -p <project> [-c <component>] [-d <dest>] <command>
 
 Options
 -------
 -c <component> Select components from the project directory.
                Multiple components may be selected by specifying multiple '-c'
                options; at least one component must be specified.
--d <dir>       The name of the directory for the rootfs.
-               Note: This is also used for the .tar and .img file names.
+-d <dest>      Destination directory in which brickstrap should store output.
```
dlech: s/should/will/

Also should probably say that this directory will be created if it does not already exist.

brickstrap.sh
```diff
@@ -40,35 +40,43 @@ function br_script_path()
...
 -p <project>   Directory which contains the brickstrap configuration (project).
                This option is required and should occur exactly once.
                Values are either a path to the project directory or the name of
                an example project shipped with brickstrap by default.
--Q <qemu>      Optional: override which QEMU binary to use for emulation of
+-I <image>     Name of the disk image file to generate (without type suffix).
```
dlech: Should make it clear that this is an optional parameter. Also, I think it should include the suffix. Someone may want `.raw` instead of `.img` or something like that.

brickstrap.sh
```diff
@@ -40,35 +40,43 @@ function br_script_path()
...
                architecture which matches the host architecture or 'native'
                then no emulator will be used. Otherwise the system is queried
                for a well known QEMU emulator for the architecture matching the
-               <qemu> value.
+               <emulator> value.
+-B <filename>  Blacklist a package file from the project. The file will be
+               ignored while determining the set of packages to install. If the
+               project does not use the PACKAGES variable this setting will
+               have no effect.
+-P <package>   Blacklist a package (name). The package will not be added to the
```
dlech: I'm not sure that I like this option. For example, a package could still be installed even if it is listed here because it is a dependency of another package.

Likewise, rather than blacklisting full package lists, it seems like we should just advocate making components more granular.

Also, if you blacklist a package and there is a hook that depends on it, you will have to blacklist the hook as well. I just see these options causing more problems than they fix and I am not looking forward to providing technical support for them.

If you really feel they are necessary, I would be OK with having them as an environment variable as sort of a back door, but I'm just not liking the idea of having blacklisting as a regular command line option.

dlech:
The rest of this looks good (other than my couple comments).

cmacq2:
> What is the use case for multiple images from a single rootfs? It seems to me that if you want a different image, you would have to build a new rootfs. I would just as soon have the (single) image not in a directory by itself.

That depends. If you have a single project which supports multiple devices you may want to generate multiple types of rootfs in a single build. For example:
 * The Raspberry partitioning layout is quite specific to work around quirks of the R.Pi bootloader.
 * Devices like the BeagleBone have a dedicated flash. You may want to use flash friendly filesystems like UBIFS or JFFS2.
 * You might want to generate both a device rootfs and a VM (e.g.: kvm+qemu) image.

Also having an `images/` directory is slightly more robust in the face of arbitrary `-I` values (edge case: `-I rootfs`, (custom) driver that simply copies the tar file, use case: deployment to NFS later on).

brickstrap-argv.sh
cmacq2: I dunno. The downside of something like Python is that it doesn't buy you much for the actual work of brickstrap which is mostly doing things like `cat`/`cp`,`ln` and globbing with `*` apart from invoking the multistrap and user-unshare scripts.

brickstrap.sh
cmacq2: I'd like to leave type suffixes to the driver. When generating multiple variations of the same rootfs image (multiple drivers) the image name should reflect which driver generated it and ideally the image suffix should reflect something about the type of disk image (e.g. `.vmdk` or `.mbr`/`.img`) generated.

Injecting that information probably breaks expectations if you let the user pick suffixes as well because then they might quite reasonably think they control the actual output name instead of just a 'template' value which information is tacked onto.

Users that are really concerned about this can write a custom deployment script that fixes this with `mv` commands.

cmacq2:
> If we are going to get into this sort of fine-grained options, I would like to have long option names. Of course, that is out of the scope of this pull request.
>
Also, I would like to choose more obscure letters for the short form of blacklisting so that we can save the more common letters (P and B) for future options where they might make more sense.

Any particular letters in mind?

> I'm not sure that I like this option. For example, a package could still be installed even if it is listed here because it is a dependency of another package.
>
>Likewise, rather than blacklisting full package lists, it seems like we should just advocate making components more granular.
>
>Also, if you blacklist a package and there is a hook that depends on it, you will have to blacklist the hook as well. I just see these options causing more problems than they fix and I am not looking forward to providing technical support for them.

The problem stems from the fact that there's a distinction between the end-user of a project configuration and its developer. The blacklisting mechanism is a tool for the end-user to selectively override some of the authors decisions: in particular if a user wishes to trim down an image somewhat. 

Essentially it is a 80/20 solution to the problems of configuration authors having a somewhat one sided focus on the needs of their own pet project, historical baggage and/or cutting corners in getting a board up and running quickly. The alternative is for users to fork the configuration or develop their own configuration but that is a rather steeper learning curve altogether.

For example your ev3dev projects pull in a lot of runtimes but the user might decide they don't really need/want a `nodejs` and don't want their time and bandwidth wasted on downloading it.

As for hooks: we could of course add a mechanism to blacklist hooks (easy!) but that is where your argument actually rings very true:

> it seems like we should just advocate

I.e.: hooks should be grouped inside a component together with their corresponding packages. That way the user simply picks their component selection appropriately and the problem is avoided entirely as well.
-----------------
Tangent:

We should advocate also that developers not use hooks at all whenever feasible because they amount to interference with the package management system - i.e. the *proper* thing to do with hooks is to package your integration bits and then specify it among the packages to install. Then you don't need hooks and you automatically have this compatibility sorted for you by the package manager. For example hooks which muck around in `policy-rc.d` shouldn't really be run at all when your init system is systemd (and vice versa: hooks that touch systemd configs shouldn't run for a sysv init system).
-------------------

Again the emphasis should be on the fact that it is an 80/20 solution to trim down some subjective bloat, but not a general mechanism for users to override the configuration developer's choices. The documentation should reflect and emphasise that.

dlech:
> Any particular letters in mind?

How about `-x` to exclude a single package and `-X` to exclude a package list file?

dlech:
This is all much fancier than I really have a need for but your reasoning is sound. I hope that you will do some advertising to encourage people to use brickstrap to benefit from all these features that you are adding. :wink:

I'm going to have to take this for a test drive, but I think you've talked me out of most of my descents here.

cmacq2:
> How about -x to exclude a single package and -X to exclude a package list file?

Sounds good, I'll update the PR accordingly.

cmacq2:
I've retouched the usage/help output to incorporate your feedback, and renamed the blacklisting short options as per your suggestion.

dlech:
I made the required variable name changes and consolidated the *-ev3dev-jessie projects into a single project. I'm very happy (and a little surprised) that I was able to do it without a single problem. It worked as expected on the first try. So great job and keep up the good work!

Issue #10, opened 2015-12-10
Allow for custom image partition scheme


BrendanSimon:
# Recoverable Partition Scheme

Currently the image creation is fixed in the brickstrap.sh script.  It would be better if this was factored out user/target to configuration, so that each build type can specify it's own partition scheme, etc.

My use case is for devices deployed in remote locations, that need to be remotely upgraded, and cannot be bricked if something goes wrong in the upgrade process.  e.g. loss of network, power drop out, etc.  In general this is solved by have multiple rootfs partitions.

## My use case

Partition | Purpose | File System | Size
---- | ---- | ---- | ----
1 | BOOT | VFAT | 32MB, 64MB
2 | ROOTFS1 | ext4 | 1GB
3 | ROOTFS2 | ext4 | 1GB
4 | DATA | VFAT, ext4 | 14GB - 30GB

## More general use case

Partition | Purpose | File System | Size
--- | --- | --- | ---
1 | BOOT | VFAT, FAT32, FAT16, FAT12 | 32MB, 48MB, 64MB
2 | ROOTFS1 | ext2, ext3, ext4 | 512MB, 1GB, 2GB
3 | ROOTFS2 | ext2, ext3, ext4 | 512MB, 1GB, 2GB
4 | DATA | VFAT, ext2, ext3, ext4 | 1GB - 30GB

Other partitioning schemes can be considered, including those with logical volumes, extended volumes, lvm, etc.

# Ping-Pong Rootfs
------------------------
When a reboot occurs, the boot loader should have the capability to fallback to the last working rootfs if the latest rootfs fails integrity checks, does not load properly or does not run to a certain state where it would set something to notify the bootloader of it's success (e.g. a u-boot env var).  This is sometimes referred to as a ping-pong rootfs system.

# Golden Rootfs
-------------------
A similar approach is to have the first partition the golden partition that never gets touched once programmed in the factory, and the second rootfs is the only one that gets upgraded.  This can have issues of upgrading an upgraded system.  i.e. upgrading a running system.  It can be solved by booting into the golden partition before doing the upgrade of the 2nd partition.  This can mean the system is down while doing the upgrade, which may not be acceptable.

An extension of the Golden Rootfs is to have a 3 rootfs partions -- one golden that never ever gets touched, and two other rootfs partitions for ping-pong.  The Golden (or Recovery/Fallback) partition is usually small and simple whose main purpose is to be able to upgrade the main rootfs partitions with valid information.

[...]

PR #29, opened 2016-01-15
Allow for custom image partition schemes.

cmacq2:
This commit provides preliminary support for custom partitioning schemes/image file types.
For discussion refer to: issue #10:
https://github.com/ev3dev/brickstrap/issues/10

This change assumes some artefacts provided by PR #28 are available:
https://github.com/ev3dev/brickstrap/pull/28
Specifically: br_image_dir() and br_image_path() functions.

This change does **not** address integration within brickstrap.sh itself (should be trivial, though).

### DIFF
--- a/brickstrap-image-drivers.sh
+++ b/brickstrap-image-drivers.sh
@@ -1,0 +1,152 @@
+#!/bin/bash
+#
+# This file is part of brickstrap.
+#
+# brickstrap - create a foreign architecture rootfs using kernel namespaces,
+#              multistrap, and qemu usermode emulation and create a disk image
+#              using libguestfs
+#
+# Copyright (C) 2016 Johan Ouwerkerk <jm.ouwerkerk@gmail.com>
+#
+
+#
+# Note: this file is not meant to be executed, source it as a library of
+# functions instead. Variables used by the functions (other than stack) are
+# namespaced using the 'BR_' or 'BRP_' prefixes. Function names are namespaced
+# similarly, using the 'br_' and 'brp_' prefixes. See docs/namespacing.md for
+# more information on namespacing in brickstrap.
+#
+
+#
+# This file provides driver functions for default image types supported by
+# brickstrap.
+#
+
+#
+# Implement a boot+root partition scheme.
+# This function creates a 'boot+root' MBR type image (type extension: img).
+# $1: path to the image file to generate.
+#
+function brp_image_drv_bootroot()
+{
+    debug "IMAGE_FILE_SIZE: ${IMAGE_FILE_SIZE}"
+    [ $# -eq 1 -a -n "$1" ] && guestfish -N \
+        "$1"=bootroot:vfat:ext4:${IMAGE_FILE_SIZE}:48M:mbr \
+        part-set-mbr-id /dev/sda 1 0x0b : \
+        set-label /dev/sda2 EV3_FILESYS : \
+        mount /dev/sda2 / : \
+        tar-in "$(br_tarball_path)" / : \
+        mkdir-p /media/mmc_p1 : \
+        mount /dev/sda1 /media/mmc_p1 : \
+        glob mv /boot/flash/* /media/mmc_p1/ : \
+
+    # Hack to set the volume label on the vfat partition since guestfish does
+    # not know how to do that. Must be null padded to exactly 11 bytes.
+    echo -e -n "EV3_BOOT\0\0\0" | \
+        dd of="$1" bs=1 seek=32811 count=11 conv=notrunc >/dev/null 2>&1
+}
+
+#
+# Convert megabytes to sectors. Sectors are assumed to be 512 bytes big.
+# $1 is size in megabytes.
+function brp_to_sector()
+{
+    #echo $(( $1 * 1024 * 1024 / 512 ))
+    echo $(( $1 * 2048 ))
+}
+
+#
+# Create a disk image with MBR partition table and 4 partitions. There are two
+# identical rootfs partitions to allow for failover and live upgrades.
+# ------------------------------------------------------------------------------
+# part | label              | mount point | fs   | size
+# ------------------------------------------------------------------------------
+#    1 | ${BOOT_PART_NAME}  | /boot/flash | VFAT | 48MB
+#    2 | ${ROOT_PART_NAME}1 | /           | ext4 | ${ROOT_PART_SIZE}
+#    3 | ${ROOT_PART_NAME}2 | /mnt/root2  | ext4 | ${ROOT_PART_SIZE}
+#    4 | ${DATA_PART_NAME}  | /var        | ext4 | ${IMAGE_FILE_SIZE} -
+#      |                    |             |      | 2 * ${ROOT_PART_SIZE} - 48MB
+# ------------------------------------------------------------------------------
+#
+# $1: path to the image file to generate.
+#
+function brp_image_drv_redundant_rootfs_w_data()
+{
+    [ $# -eq 1 -a -n "$1" ] && guestfish -N "$1"=disk:${IMAGE_FILE_SIZE} -- \
+        part-init /dev/sda mbr : \
+        part-add /dev/sda primary 0 $(brp_to_sector 48) : \
+        part-add /dev/sda primary $(brp_to_sector 48) $(brp_to_sector ${ROOT_PART_SIZE}) : \
+        part-add /dev/sda primary $(brp_to_sector ${ROOT_PART_SIZE}) $(brp_to_sector $(( 2 * ${ROOT_PART_SIZE} ))) : \
+        part-add /dev/sda primary $(brp_to_sector $(( 2 * ${ROOT_PART_SIZE} ))) -1 : \
+        part-set-mbr-id /dev/sda 1 0x0b : \
+        mkfs fat /dev/sda1 : \
+        set-label /dev/sda1 ${BOOT_PART_NAME} : \
+        mkfs ext4 /dev/sda2 : \
+        set-label /dev/sda2 ${ROOT_PART_NAME}1 : \
+        mkfs ext4 /dev/sda3 : \
+        set-label /dev/sda3 ${ROOT_PART_NAME}2 : \
+        mkfs ext4 /dev/sda4 : \
+        set-label /dev/sda4 ${DATA_PART_NAME} : \
+        mkdir-p /boot/flash : \
+        mount /dev/sda1 /boot/flash : \
+        mount /dev/sda2 / : \
+        mkdir-p /mnt/root2
+        mount /dev/sda3 /var : \
+        mkdir-p /var : \
+        mount /dev/sda4 /var : \
+        tar-in "$(br_tarball_path)" / : \
+        umount /boot/flash : \
+        umount /var : \
+        glob cp-a /* /mnt/root2/ : \
+
+}
+
+#
+# Sanity checks configuration variables for the redundant rootfs + data image
+# type. If variables aren't defined a default is set.
+#
+# Variables:
+# BOOT_PART_NAME: Label of boot partition. Default: BOOT
+# ROOT_PART_NAME: Label of root partition. Default: ROOTFS
+# DATA_PART_NAME: Label of root partition. Default: DATA
+# IMAGE_FILE_SIZE: The size of the entire image file. Default: 3800M
+#
+function brp_image_drv_check_redundant_rootfs_w_data()
+{
+
+    BOOT_PART_NAME=${BOOT_PART_NAME:-BOOT}
+    ROOT_PART_NAME=${ROOT_PART_NAME:-ROOTFS}
+    DATA_PART_NAME=${DATA_PART_NAME:-DATA}
+    IMAGE_FILE_SIZE=${IMAGE_FILE_SIZE:-3800M}
+
+    [ ${#BOOT_PART_NAME} -gt 11 ] && \
+        fail "BOOT_PART_NAME cannot be more than 11 characters."
+
+    # see https://en.wikipedia.org/wiki/Label_%28command%29
+    echo $BOOT_PART_NAME | egrep -q '^[A-Z0-9_-]*$' || \
+        fail "BOOT_PART_NAME contains invalid characters"
+
+    # appending 1/2 to ROOT_PART_NAME, so 16 characters total
+    [ ${#ROOT_PART_NAME} -gt 15 ] && \
+        fail "ROOT_PART_NAME cannot be more than 15 characters."
+    echo $ROOT_PART_NAME | egrep -q '^[a-zA-Z0-9_-]*$' || \
+        fail "ROOT_PART_NAME contains invalid characters"
+
+    [ ${#DATA_PART_NAME} -gt 16 ] && \
+        fail "DATA_PART_NAME cannot be more than 16 characters."
+    echo $DATA_PART_NAME | egrep -q '^[a-zA-Z0-9_-]*$' || \
+        fail "DATA_PART_NAME contains invalid characters"
+}
+
+#
+# Registers the default drivers. To be invoked by brickstrap early before doing
+# any option parsing etc.
+#
+function brp_image_drv_register_defaults()
+{
+    br_register_image_type bootroot brp_image_drv_bootroot \
+        img # no validation required for bootroot, yet
+    br_register_image_type redundant brp_image_drv_redundant_rootfs_w_data \
+        img brp_image_drv_check_redundant_rootfs_w_data
+}
+
--- a/brickstrap-image.sh
+++ b/brickstrap-image.sh
@@ -1,0 +1,210 @@
+#!/bin/bash
+#
+# This file is part of brickstrap.
+#
+# brickstrap - create a foreign architecture rootfs using kernel namespaces,
+#              multistrap, and qemu usermode emulation and create a disk image
+#              using libguestfs
+#
+# Copyright (C) 2016 Johan Ouwerkerk <jm.ouwerkerk@gmail.com>
+#
+
+#
+# Note: this file is not meant to be executed, source it as a library of
+# functions instead. Variables used by the functions (other than stack) are
+# namespaced using the 'BR_' or 'BRP_' prefixes. Function names are namespaced
+# similarly, using the 'br_' and 'brp_' prefixes. See docs/namespacing.md for
+# more information on namespacing in brickstrap.
+#
+
+#
+# Default image type 'driver'. This function simply delegates to the current
+# brickstrap default.
+# $1: path to the image file to generate.
+#
+function brp_image_default_driver()
+{
+    brp_image_drv_bootroot "$1"
+}
+
+#
+# Default image type configuration validator. This function simply delegates to
+# the current brickstrap default.
+#
+function brp_validate_default_image_config()
+{
+    true # there is no validation required for the simple bootroot driver, yet
+}
+
+#
+# Displays an error message if a image type driver failed to create an image
+# file successfully. Returns the given status code.
+# $1: the status code returned by the driver function
+# $2: the name of the partitioning/imaging scheme implemented in the disk image
+# $3: the file path of the disk image
+#
+function brp_image_err_msg()
+{
+    error "Unable to create image: '$3'.
+Driver '$2' ($(brp_get_image_type_driver "$2")) returned status code: $1"
+    return $1
+}
+
+#
+# Create a disk image file (type). This function looks up the
+# relevant driver and invokes it with the given path. Before creating the image
+# this function checks it does not already exist or, if it does, that it may be
+# overwritten.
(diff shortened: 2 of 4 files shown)


dlech:
Looks good other than as I mentioned in #29, I'm not seeing a need for creating multiple image files at one time.

dlech:
Please explain how you see this image registry working. Also, how does the new `-I` option work when you have multiple image files?

I'm envisioning having a folder called `image-drivers` that contains individual files for each driver. The driver used by a project is specified by setting a variable in the `config` file of one of the components. brickstrap searches first `$(br_project_dir)/image-drivers`, then `$(br_script_path)/image-drivers` for a matching driver. This way you don't have to mess with registering any functions. The file will just be an executable script that creates the image.

cmacq2:
> Also, how does the new -I option work when you have multiple image files

Drivers are registered with a 'name' that identifies the driver as well as a file type suffix. Names are unique. The disk image file name is generated from the basename of `-I` as follows:

`<basename>-<driver name>.<file type suffix>`

So the necessary identifying information is simply appended to the basename specified with `-I`.

> Please explain how you see this image registry working. 

In a follow up integration PR, we introduce a commandline option `-F` or something to select rootfs type. This will mediate between project defaults and end-user choice.

From the configuration developer perspective:

 * You may split support for different image types over components (say, `fs-beagle`, `fs-rpi`, `fs-kvm`).
 * Each component that needs support for a custom image type contains its own `custom-image.sh` file.
 * In the `custom-image.sh` you do something like this:
```
#
# create custom image type:
# $1: disk image file to write.
#
function my_custom_img_driver_func()
{
# ... implement it here...
}

br_register_image_type "my_driver_name" my_custom_img_driver_func "img-file-suffix"
```
 * Then in a `config` file you can adjust `IMAGE_TYPES` environment variable to include "my_driver_name". Alternatively you can require the user select it with `-F my_driver_name`. `IMAGE_TYPES` allows the project configuration to override the default image type when no image type has been selected through `-F`.
 * As a project developer you should document support for additional image types and mention what components (not to mention programs on the host) are required for each additional driver.

From the end-user perspective you can specify the exact set of image types to generate, provided corresponding drivers are available either from brickstrap defaults or from the component selection,  using the `-F` option.

dlech:
It's getting late and I'm about to call it quits, so I haven't read your previous comment yet. However, I wanted to let you know that I just pushed a [branch with some image drivers](https://github.com/ev3dev/brickstrap/tree/image-drivers) so that you can see what I am hoping for in an image driver. Feel free to make comments on it as to how you see it fitting into what you are proposing here.

dlech:
One thing I would like to do is call the `validate` function early in the process to catch as many configuration errors as possible before running through the full build.

cmacq2:
> It's getting late and I'm about to call it quits, so I haven't read your previous comment yet. However, I wanted to let you know that I just pushed a branch with some image drivers so that you can see what I am hoping for in an image driver. Feel free to make comments on it as to how you see it fitting into what you are proposing here.

The thing I like:

 * Support for custom settings and validation thereof. I think this existing PR could benefit from incorporating a mechanism for that.

The aspects which I think just don't work right at the moment, and are better addressed in this PR already:

 * Copy-paste code. This model encourages copy-paste code.
 * This also means potentially copy-paste bugs, reinventing of wheels, doing work twice.
 * In particular: your drivers all seem to want to create the same file (`<basename>-bootroot.img`)
 * Which precludes you from generating multiple rootfs files at the same time.
 * All your drivers duplicate the check for whether this file exists, and whether `-f` was specified.
 * Inline code instead of functions. This means you have to execute the script, instead of sourcing. This also means you can't reuse functions that depend on state, efficiently. In particular your calls to `$(br_image_path)` require that the `BR_IMAGE_BASE_NAME` and `BR_DESTDIR` variables are set. There's nothing in your code to receive those from brickstrap itself... Likewise `BR_FORCE` is also not available.
 * The same problem actually prevents your settings feature from working correctly (in the face of arbitrary variable names).
 * Similarly it doesn't gel well with the existing components mechanism which is set up to source scripts rather than execute them.

cmacq2:
I've updated my PR to provide necessary integration bits as well.

Additionally I added support for validation of custom configuration options (register additional validator functions), and I've also integrated one of your default drivers, so you can see how that should work now. See brickstrap-image-drivers.sh for the example of how I integrated the "redundant rootfs + data" image type.

brickstrap-image.sh
```diff
@@ -0,0 +1,221 @@
...
+    fi
+    if [ -n "$IMAGE_TYPES" ]; then
+        brp_create_image_types "$IMAGE_TYPES" "$BR_IMAGE_TYPES" || \
+            BRP_IMAGE_RETURN_CODE=$?
+    fi
+    if [ -z "$BR_IMAGE_TYPES" -a -z "$IMAGE_TYPES" ]; then
+        brp_create_image_type img || BRP_IMAGE_RETURN_CODE=$?
+    fi
+
+    return $BRP_IMAGE_RETURN_CODE
```
dlech: If you are building more than one image and the first one fails and later ones succeed, won't this function still return success? It seems like if one fails we should just be calling `fail` and stopping brickstrap altogether.

brickstrap-image-drivers.sh
```diff
@@ -0,0 +1,151 @@
...
+#
+# $1: path to the image file to generate.
+#
+function brp_image_drv_redundant_rootfs_w_data()
+{
+    [ $# -eq 1 -a -n "$1" ] && guestfish -N "$1"=disk:${IMAGE_FILE_SIZE} -- \
+        part-init /dev/sda mbr : \
+        part-add /dev/sda primary 0 $(brp_to_sector 48) : \
+        part-add /dev/sda primary $(brp_to_sector 48) $(brp_to_sector ${ROOT_PART_SIZE}) : \
+        part-add /dev/sda primary 0 $(brp_to_sector ${ROOT_PART_SIZE}) $(brp_to_sector $(( 2 * ${ROOT_PART_SIZE} ))) : \
```
dlech: My bad. The `0` should be removed on this line.

brickstrap-image.sh
cmacq2: > If you are building more than one image and the first one fails and later ones succeed, won't this function still return success? 

It shouldn't. `brp_create_image_types` should return non-zero if any of the types fail (it uses a different variable internally so it is safe).  The `||` means that `BRP_IMAGE_RETURN_CODE` is set only on failure. So the end result should be that the exit code will correspond to the one from the last image to fail.

brickstrap-image-drivers.sh
cmacq2: Will adjust.

brickstrap-image.sh
dlech: got it.

cmacq2:
> It seems like if one fails we should just be calling fail and stopping brickstrap altogether.

I think it is preferable to let the create-image step continue as much as it can. During development of a project configuration it may be that you accidentally break more than one image type, so it seems preferable to have a somewhat more comprehensive report (via the error messages) than just failing on the first bad image. 

Also, it makes it slightly easier for an end-user to work around a broken image type: even if one image build fails, another might succeed and present a viable interim alternative.

dlech:
> <basename>-<driver name>.<file type suffix>

I would really rather not have the driver name part of the file name if I am creating only one image file.

dlech:
To keep things consistent with the images I have been publishing, the image name should just be `<dest>.img` where `<dest>` is the `-d` option.

cmacq2:
> I would really rather not have the driver name part of the file name if I am creating only one image file.

Sure, that can be done. I don't really see why, though.

> To keep things consistent with the images I have been publishing, the image name should just be...

You could also execute brickstrap differently, simply specify -I. If we agree to special case the single-driver case, then it should be sufficient to specify `-I` the way you want it...

dlech:
How about the case when you want to make multiple images using the same driver with different parameters, for example IMAGE_SIZE? Rather than trying to handle such cases in brickstrap, I propose that we just advocate the following...

Brickstrap only creates one image file when you run `brickstrap ... all`. The driver that is used is specified in the `config` file in the project. The resulting file name is `<dest>.<ext>` where `<dest>` is the `-d` option and `<ext>` is determined by the image driver.

If you want to override the default settings of the driver, then use the corresponding environment variables. e.g. `IMAGE_FILE_SIZE=1800M brickstrap ... all`.

If you want to override the driver, use the `-l` option. e.g. `brickstrap ... -l beaglebone all`

If you want to override the basename of the image file, use the `-I` option. e.g. `brickstrap ... -I my-custom all` (results in `my-custom.img`)

If you want to build additional images, then use the `create-image` command after the `all` command.

    brickstrap ... -I my-project-4GB all
    IMAGE_FILE_SIZE=1800M brickstrap ... -I my-project-2GB create-image
    IMAGE_FILE_SIZE=900M brickstrap ... -I my-project-1GB create-image
    brickstrap ... -l beaglebone -I my-project-beaglebone-4GB create-image

For now, you can abuse the `create-report` hook if you need to automate creating more than one image.

cmacq2:
> How about the case when you want to make multiple images using the same driver with different parameters, for example IMAGE_SIZE? Rather than trying to handle such cases in brickstrap, I propose that we just advocate the following...

Uh: you can do so right now using the exact same create-image workflow you suggest already. There is no extra handling or support in brickstrap required for that, and I certainly do not propose we add it (I don't see what this would buy us, see next paragraph).

Besides: how often would you want to do *that* versus wanting to make different image types? I'd think that between those two cases the scales would tip in favour of different types over sizes -- mainly because you want generally either do not statically fix the size of the image or you have a very particular target in mind anyway (so only one specific size).

> The resulting file name is <dest>.<ext> where <dest> is the -d option and <ext> is determined by the image driver.

Let's keep the inferred default file name for the image as it is in this PR. I think the alternative has a few drawbacks:

 * File extensions are not necessarily unique. Even if they were, they are also cryptic in the sense that it's not at all obvious what the actual partitioning layout of the rootfs is inside. Even if that were not the case, it's still not immediately obvious what `-l` option was passed to build the image.
 * The timestamp embedded in the file name is actually kind of useful: it permits an incremental build workflow that preserves older images and it makes it immediately obvious when an image was built even if modification timestamps are bogus (e.g. downloaded files).

-----------------

I think my main objection is that the proposal doesn't make the brickstrap code significantly less complicated (all it really would amount to is removing two outer loops that invoke two particular functions so you can invoke those two functions with fixed arguments directly). At the same time it also does not make brickstrap any easier to use, having to fiddle with options.

So I don't see what we gain by reverting to generating only a single fixed image, albeit one selected via commandline options.

cmacq2:
About yocto specifically: [this section of its manual](http://www.yoctoproject.org/docs/1.8/ref-manual/ref-manual.html#image-generation-dev-environment) explains how image generation works with yocto/open-embedded. 

TL;DR summary is: you set up the variable [`IMAGE_FSTYPES`(http://www.yoctoproject.org/docs/1.8/ref-manual/ref-manual.html#var-IMAGE_FSTYPES) in a local build config

dlech:
> Uh: you can do so right now using the exact same create-image workflow you suggest already.

The point that I am trying to make is that if this already works, why are we inventing a new method to do the same thing? The complexity is not in the implementation, but rather in trying to explain "if you want to make more than one image, you can do it this way, but it only works in some cases otherwise, you have to do it this other way". Just saying "do it this way" is much simpler.

> how often would you want to do that versus wanting to make different image types?

Always. I can't think of a single real-life use case for me personally (distributing images for EV3, RPi1/2 and BeagleBone) that could not be accomplished by using the default image driver and environment variables (that would be set in the `config` of the appropriate component directory). In the example drivers I gave earlier, the two cases that I would actually use are the default and the beaglebone. However, the beaglebone is identical other than the two extra `dd`s for the bootloader, so it doesn't really make sense to make it a separate driver. I would just make it part of the default driver and put if statements around the beaglebone specific stuff so that it is ignored unless the variables are set. In real life, I wouldn't actually generate various sized images, I was just giving that as a simple example of using a variable in the context of the existing default image driver.

Although the full truth is that the answer is neither. I don't have a single real-life case where I would want to produce more than one image from a given rootfs.

> Let's keep the inferred default file name for the image as it is in this PR.

That's fine as long as there is some way I can have full control over the entire basename of the file name (not including the file extension).

> The timestamp embedded in the file name is actually kind of useful

By default, (i.e. the `-d` option is omitted), the destination directory should have the timestamp in the directory name already. So, by using `<dest>` as the image basename, it should have the timestamp in it already.

---

Offtopic: The reason I keep pushing back so hard about these theoretical use cases is exactly that - they are theoretical. I maintain brickstrap because it meets a real need for me (distributing images for the ev3dev.org project). If others come along with real needs and want to modify brickstrap in a way that can meet their need as well without breaking my needs, then great. However, I am quickly getting tired of spending my time thinking about what someone *might* want to do instead of working on something that actually meets a real need. @BrendanSimon has a real need for something other than the default image driver, so if we can make some sort of hook for him to latch on to, then great, but I expect him to do most of the work for actually implementing the driver to meet his needs. I would like to keep brickstrap as simple and minimalistic as possible. If you are interested in something more all-encompassing, maybe checkout [ELBE](https://github.com/Linutronix/elbe) that I heard about recently on the emdebian mailing list.

These are great ideas that you have, I just can't justify spending the time it is taking me to keep up at this rate with your innovations and really understand what you are doing. It is unfortunate, but that is the way it is.

dlech:
Or if you want to reverse the roles where you are the maintainer and I am a contributor that is an option too. :smile:

cmacq2:
OK: that is fair enough.

Going forward, it seems that the basic functionality of the PR is agreed on but that image generation should be restricted to a single image.

> By default, (i.e. the -d option is omitted), the destination directory should have the timestamp in the directory name already. So, by using <dest> as the image basename, it should have the timestamp in it already.

This is not currently the case: by default it is actually `$(pwd)`. I think that is actually preferable in the single image case because it means you can issue multiple `create-image` commands without specifying 'redundant' options or risking the default getting in your way by 'changing' due to elapsed time.

I'll readjust the PR to take your feedback into account and rework the `IMAGE_TYPES` variable to a `DEFAULT_IMAGE_TYPE` (to be used as a default only).

cmacq2:
Ok, so how does this version look which reduces the number of images being generated in a single brickstrap back down to one and ensures you have full control over the image base name (sans suffix) through `-I`.

By default an image name looks like this: `<destdirname>-<timestamp>-<driver>.<suffix>`
Using the `-I <custom_name>` option it becomes: `<custom_name>.<suffix>`

brickstrap-image.sh
```diff
@@ -0,0 +1,202 @@
...
+
+#
+# Look up the driver function for a given image type name.
+# $1: the name of the partitioning/imaging scheme implemented in the disk image
+#
+function brp_get_image_type_driver()
+{
+    [ $# -eq 1 -a -n "$1" ] && case "$1" in
+        default)
+            echo -n brp_image_default_driver
```
dlech: Are you missing `;;` here? I just tried running brickstrap from this branch and got

    brickstrap-image.sh: line 90: syntax error near unexpected token `)'

Works after I add `;;`

brickstrap-image.sh
```diff
@@ -0,0 +1,202 @@
...
+        *)
+            eval echo -n "\$BRP_IMG_DRV_REGISTRY_$1"
```
dlech: `;;`

brickstrap-image.sh
```diff
@@ -0,0 +1,202 @@
...
+
+#
+# Look up the file type extension for a given image type name.
+# $1: the name of the partitioning/imaging scheme implemented in the disk image
+#
+function brp_get_image_type_extension()
+{
+    [ $# -eq 1 -a -n "$1" ] && case "$1" in
+        default)
+            echo -n img
```
dlech: `;;`

brickstrap-image.sh
```diff
@@ -0,0 +1,202 @@
...
+        *)
+            eval echo -n "\$BRP_IMG_EXT_REGISTRY_$1"
```
dlech: `;;`

brickstrap-image.sh
```diff
@@ -0,0 +1,202 @@
...
+
+#
+# Look up the validator function to check image configuration parameters.
+# $1: the name of the partitioning/imaging scheme implemented in the disk image
+#
+function brp_get_image_cfg_validator()
+{
+    [ $# -eq 1 -a -n "$1" ] && case "$1" in
+        default)
+            echo -n brp_validate_default_image_config
```
dlech: `;;`

brickstrap-image.sh
```diff
@@ -0,0 +1,202 @@
...
+        *)
+            eval echo -n "\$BRP_IMG_CFG_VALIDATOR_$1"
```
dlech: `;;`

brickstrap-image-drivers.sh
```diff
@@ -0,0 +1,151 @@
...
+        mount /dev/sda1 /boot/flash : \
+        mount /dev/sda2 / : \
+        mkdir-p /mnt/root2
+        mount /dev/sda3 /var : \
+        mkdir-p /var : \
+        mount /dev/sda4 /var : \
+        tar-in "$(br_tarball_path)" / : \
+        umount /boot/flash : \
+        umount /var : \
+        glob cp-a /* /mnt/root2/ : \
```
dlech: Since there is a `\` at the end of the line here, it is important that the next line is blank. It's been a while since I messed with this, but I remember guestfish being weird, so I don't think removing the trailing  `: \` is an option.

brickstrap.sh
```diff
@@ -668,9 +648,10 @@ function brp_run_command()
 
 function brp_run()
 {
-    brp_parse_cli_options "$@" && \
-    brp_validate_cli_options "$BRP_CMD" "$BRP_CMD_ARG" && \
-    brp_init_env && \
+    brp_image_drv_register_defaults
```
dlech: Currently, brickstrap is exiting without an error message somewhere in this function.

brickstrap-image.sh
```diff
@@ -0,0 +1,202 @@
...
+# $1: the name of the partitioning/imaging scheme implemented in the disk image
+# $2: name of the function which will create the image (type), taking the image
+#     file path as argument.
+# $3: the image file type extension, without leading dot.
+# $4: optional: a validator function to check parameters intended for the image
+#     driver function in the project configuration.
+#
+function br_register_image_type()
+{
+    [ $# -eq 3 -a -n "$1" -a -n "$2" -a -n "$3" ] && \
```
dlech: I changed this line to give an error message if we have bad parameters.
    
    [ $# -eq 3 -a -n "$1" -a -n "$2" -a -n "$3" ] || \
        fail "Incorrect number of parameters for br_register_image_type"

dlech: Nope. That's not how to fix it.

dlech: How about this?

    
    [ $# -eq 3 -o $# -eq 4 ] && [ -n "$1" -a -n "$2" -a -n "$3" ] || \
        fail "Bad parameters for br_register_image_type"

cmacq2: No it has to be `$# ge 3` now because we now accept an optional 4th param (validator function).

dlech: `[ $# -eq 3 -o $# -eq 4 ]` does essentially the same thing, except it will fail if there are 5 parameters.

cmacq2:
I'll update the PR to include the fixes found.

dlech:
You may want to wait a bit. I still can't get brickstrap to run.

cmacq2:
> I still can't get brickstrap to run.
At what point does it fail? I'm currently doing a test run and it proceeds happily to download stage with options: `-I foo -p /full/path/to/my/project -c . all`

dlech:
Somewhere in `brp_parse_cli_options`. Haven't figured it out exactly. My command line is 
   
    brickstrap.sh -p ev3dev-jessie -c base -c debian -c ev3 all -f

dlech:
Found the problem. #35

brickstrap-image.sh
```diff
@@ -0,0 +1,208 @@
...
+# $1: the name of the partitioning/imaging scheme implemented in the disk image
+# $2: name of the function which will create the image (type), taking the image
+#     file path as argument.
+# $3: the image file type extension, without leading dot.
+# $4: optional: a validator function to check parameters intended for the image
+#     driver function in the project configuration.
+#
+function br_register_image_type()
+{
+    [ $# -ge 3 -a -n "$1" -a -n "$2" -a -n "$3" ] && \
```
dlech: Since this is a public function, I would still rather see an error message here rather than failing silently if the parameters are bad.

cmacq2: Agreed. Will update the PR.

cmacq2:
`br_register_image_type` now emits an error message with `fail` if arguments are missing or empty.

dlech:
Thanks for compromising with me.

PR #37, opened 2016-01-23
Tighten validation of project directories.

cmacq2:
With the move to a unified projects/ hierarchy for default/example projects shipped with the brickstrap code, the validation logic for the -p switch needs a corresponding update to reflect this.


cmacq2:
Will merge this with other enhancements, closing.

Issue #34, opened 2016-01-23
delete command does not delete the <dest> directory


dlech:
Now that all of the generated files live under one directory, the `delete` command should `rm -rf` that directory.

dlech:
If the `-d` option is not specified, the dest directory is `$(pwd)`. In that case, `rm -f *` would make more sense. @cmacq2, what is the best way to tell if `-d` was specified or not?

cmacq2:
That's actually the wrong approach. What you should do is check whether `$(pwd)` is at or underneath `$(br_dest_dir)`.  Because if you delete the current working directory the shell can get cranky...

cmacq2:
Since PR #44  is now merged, this issue can probably be closed.

Issue #32, opened 2016-01-22
Component inheritance.


dlech:
Now that I have set up the ev3dev-jessie projects as a single project, I see that it would be benificial for components to be able to include other components. I had to [leave myself some notes](https://github.com/ev3dev/brickstrap/tree/master/projects/ev3dev-jessie#description-of-components) about how the different components work and how to mix and match them.  It would be much easier make the includes part of the board definition so you don't have to type them every time. The inheritance looks like this...

    ev3: base, debian
    rpi-base: base
    rpi1: rpi-base, raspbian
    rpi2: rpi-base, debian

Obviously, this is not actually how it would be implemented. But, it would allow be to just specify the one component instead of having to know which 3 to choose.

cmacq2:
The easiest way to do this is to permit the `config` file to pull in additional components by setting a `EXTRA_COMPONENTS` variable and making sure that iterate components does two loops:
 * One for `BR_COMPONENTS`
 * One over `EXTRA_COMPONENTS` as well.

You probably want a helper function to delegate to for the actual loop logic, and then call it twice appropriately from the current looping function.

You would need to check you do not include the same component twice through `EXTRA_COMPONENTS` (that is you need to silently omit duplicates pulled in via `EXTRA_COMPONENTS`).

... Basically it would look like an almost exact copy of the logic that PR #29 used when it still supported multiple images.

PR #39, opened 2016-01-23
Minor enhancements/fixes.

cmacq2:
- Fix: delete destination directory if possible & sensible. See issue #34
   https://github.com/ev3dev/brickstrap/issues/34
 - Make it easier to detect in brickstrap code if '-d' option was specified by moving the default destination directory value to its own variable.
 - Enhancement: permit component inheritance through specifying EXTRA_COMPONENTS in config. See issue #32
   https://github.com/ev3dev/brickstrap/issues/32
 - Tighten validation of project directories.
   With the move to a unified projects/ hierarchy for default/example projects shipped with the brickstrap code, the validation logic for the -p switch needs a corresponding update to reflect this.

### DIFF
--- a/brickstrap-components.sh
+++ b/brickstrap-components.sh
@@ -18,6 +18,126 @@
 #
 
 #
+# Extract the component from a file inside a project structure.
+# This works only for files inside the component directory itself.
+# $1: path to translate back to its component name
+#
+function brp_path_to_component()
+{
+    [ $# -eq 1 -a -n "$1" ] && dirname "${1##$(br_project_dir)/}"
+}
+
+#
+# Read an include file which lists components to add, one per line.
+# $1: the file to read.
+#
+function brp_read_include_file()
+{
+    # Add component name to list of extra components.
+    # Avoid duplicate component names being added to BRP_EXTRA_COMPONENTS
+    # This could otherwise happen if components from BR_COMPONENTS are also
+    # listed in parsed include files.
+    #
+    BRP_IMPORTED_COMPONENT="$(brp_path_to_component "$1")"
+    brp_is_component_included "$BRP_IMPORTED_COMPONENT" || \
+    if [ -z "$BRP_EXTRA_COMPONENTS" ]; then
+        BRP_EXTRA_COMPONENTS="'$BRP_IMPORTED_COMPONENT'"
+    else
+        BRP_EXTRA_COMPONENTS="$BRP_EXTRA_COMPONENTS '$BRP_IMPORTED_COMPONENT'"
+    fi
+
+    # Read include file
+    while IFS='' read -r BRP_CUR_LINE || [ -n "$BRP_CUR_LINE" ]; do
+        case "$BRP_CUR_LINE" in
+        \#*|\;*) # permit comments: lines starting with # or ; are ignored.
+        ;;
+        *)
+            # avoid redundant spaces, i.e.  empty lines are ignored.
+            # avoid adding duplicate component names
+            if [ -z "${BRP_CUR_LINE##/}" ] || \
+                brp_is_component_included "$BRP_CUR_LINE" || \
+                br_is_component_inherited "$BRP_CUR_LINE"; then
+                continue
+            else
+                brp_validate_component_name "$BRP_CUR_LINE"
+                if [ -z "$BRP_INCLUDES" ]; then
+                    BRP_INCLUDES="'${BRP_CUR_LINE##/}'"
+                else
+                    BRP_INCLUDES="$BRP_INCLUDES '${BRP_CUR_LINE##/}'"
+                fi
+            fi
+        ;;
+        esac
+    done < "$1"
+}
+
+function br_is_component_selected()
+{
+    [ $# -eq 1 -a -n "${1##/}" ] && \
+        echo -n "$BR_COMPONENTS" | fgrep -q "'${1##/}'"
+}
+
+function br_is_component_inherited()
+{
+    [ $# -eq 1 -a -n "${1##/}" ] && \
+        echo -n "$BRP_INCLUDES" | fgrep -q "'${1##/}'"
+}
+
+function brp_is_component_imported()
+{
+    [ $# -eq 1 -a -n "${1##/}" ] && \
+        echo -n "$BRP_EXTRA_COMPONENTS" | fgrep -q "'${1##/}'"
+}
+
+function brp_is_component_included()
+{
+    br_is_component_imported "$1" || brp_is_component_selected "$1"
+}
+
+function brp_is_new_include_file()
+{
+    if brp_is_component_included "$(brp_path_to_component "$1")"; then
+        return 1
+    elif br_is_component_inherited "$(brp_path_to_component "$1")" && \
+        [ -r "$1" ]; then
+        return 0
+    else
+        return 1
+    fi
+}
+
+#
+# Import extra components by reading include file. This function will
+# iteratively import include files (resolving recursive include structures).
+# Importing terminates once there are no 'new' readable include files left to
+# process.
+#
+function brp_import_extra_components()
+{
+    # for the first round of imports all files are 'new' by definition.
+    BRP_IMPORT_CHECK=-r
+    BRP_INCLUDES=""
+    while br_list_paths include $BRP_IMPORT_CHECK >/dev/null; do
+        br_for_each_path "$(br_list_paths include $BRP_IMPORT_CHECK)" \
+            brp_read_include_file || fail "Failed to process 'include' files"
+        BRP_IMPORT_CHECK=brp_is_new_include_file
+    done
+    BRP_EXTRA_COMPONENTS="$BRP_INCLUDES"
+    debug "Selected components: $BR_COMPONENTS"
+    debug "Inherited components: $BRP_INCLUDES"
+}
+
+
+#
+# Look up path to the root directory of example/default projects shipped with
+# brickstrap.
+#
+function brp_default_projects_tree()
+{
+    echo -n "$(br_script_path)/projects"
+}
+
+#
 # Checks that a project path doesn't map to a 'reserved' brickstrap directory.
 # Reserved brickstrap directories are directories underneath $(br_script_path)
 # which are part of brickstrap source, tests or documentation as opposed to the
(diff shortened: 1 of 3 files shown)


brickstrap.sh
```diff
@@ -600,6 +601,21 @@ function brp_delete_all() {
...
     rm -rf "$(br_image_dir)"
+    BRP_PWD=$(pwd)
+
+    # if the current working directory is at or underneath the destination
+    # directory, do the safe thing: don't rm -rf, warn about it instead.
+    if [ "${BRP_PWD##$(br_dest_dir)}" != "$BRP_PWD" ]; then
+        warn "Not removing output destination directory: '$(br_dest_dir)'
+To fix this manually, try: rm -rf '$(br_dest_dir)'"
+    # if there are unexpected/stray files left over in destination directory
+    # do the safe thing: don't rm -rf, but log an error message instead.
```
dlech: Actually, I do want it to delete stray files, like the report created by my create-report.sh hook.

cmacq2: Good point. I'll remove that particular check.

cmacq2:
Rebased to latest master.

dlech:
I'm getting `[ fail] Unable to locate a readable 'multistrap.conf'` when running 

    .../brickstrap.sh -p ev3dev-jessie -c base -c debian -c ev3 -d ev3 delete

Presumably this has something to do with the change in `brp_iterate_components`

dlech:
Hmm... It's not even entering the `br_list_paths` function at line 286 in `brickstrap.sh`

       br_list_paths "multistrap.conf" -r >/dev/null || \
        fail "Unable to locate a readable 'multistrap.conf'"

dlech:
Nevermind, the `>/dev/null` is hiding debug statements

cmacq2:
Presumably  that is because we check before we load `config`, meaning that it cannot be inherited. Let's fix that (simply move it until after revalidation).

dlech:
Working for me now. One more request. Instead of `EXTRA_COMPONENTS`, could we call it `INCLUDE_COMPONENTS`?

dlech:
I also tried adding the line `EXTRA_COMPONENTS="debian base"` to `ev3dev-jessie/ev3/config` and it does not seem to including the extra components.

dlech:
It looks like the multistrap.conf is combined from the EXTRA_COMONENTS, so this code is probably working as designed. Must have uncovered another bug.

dlech:
OK.  I think the config files are not being combined properly when using EXTRA_COMPONENTS. Maybe includes/extra components should be in a separate file?

cmacq2:
> OK. I think the config files are not being combined properly when using EXTRA_COMPONENTS. Maybe includes/extra components should be in a separate file instead of as a variable in config?

No that is by design. The design is a single top level `config` pulls in `EXTRA_COMPONENTS` it depends on.

So you cannot import additional `config` files. Imported components also cannot include additional `EXTRA_COMPONENTS` for that reason.

-------------------

And I'm not sure we should attempt to 'fix' that either. Right now we can be reasonably certain that we catch invalid `EXTRA_COMPONENTS` and because we don't do recursive imports we also do not need to track which components were imported thus far beyond the global list of `EXTRA_COMPONENTS`.

... Why exactly would you need to import extra `config` files anyway? Just how granular do you split your settings, given that the user is supposed to select feature components directly via the `-c` switch?

dlech:
> Why exactly would you need to import extra config files anyway?

So that there is not duplicated configuration in each top-level component.

My component tree is: (`<component>: <depends on>`)

    # board configurations - these are the 4 top-level components
    ev3: base debian
    evb: base debian
    rpi1: rpi-base raspbian
    rpi2: rpi-base debian
    # internal components that cannot be arbitrary used with other components
    rpi-base: base
    base:
    debian: 
    raspbian

I plan to have other components later that can be arbitrarily mixed in. But there are [special rules](https://github.com/ev3dev/brickstrap/blob/master/projects/ev3dev-jessie/README.md) that have to be followed for the existing components. I am looking to hide these details so you don't have to look at the readme every time to figure out which 3 components you have to use to get a valid image.

So, I have [base/config](https://github.com/ev3dev/brickstrap/blob/master/projects/ev3dev-jessie/base/config) that has the items shared by all boards, and [debian/config](https://github.com/ev3dev/brickstrap/blob/master/projects/ev3dev-jessie/debian/config) and [raspbian/config](https://github.com/ev3dev/brickstrap/blob/master/projects/ev3dev-jessie/raspbian/config) to sort out differences in the package sources and finally [ev3/config](https://github.com/ev3dev/brickstrap/blob/master/projects/ev3dev-jessie/ev3/config) and [rpi-base/config](https://github.com/ev3dev/brickstrap/blob/master/projects/ev3dev-jessie/rpi-base/config) because I have a mix of armel and armhf. To use EXTRA_COMPONENTS, I also have to create a new `rpi1/config` and `rpi2/config`. Basically, I have a reason to have a `config` in every component.

cmacq2:
Well we could have a `include` file that is meant for the `EXTRA_COMPONENTS` setting specifically. I'll adjust the PR accordingly and I'll rename the variable to `INCLUDE_COMPONENTS`.

As for how your components inherit other components: that's inherently recursive in nature. I'll need to rework the PR a bit for that.

Oddly enough: at least two of your `config` files are basically what a previous PR of mine wanted to introduce the `BR_ARCH` variable (a switch with Debian architecture name) for, though I suppose the difference is that you want to hide the detail of the architecture instead of providing a template config that just needs the architecture 'plugged in'...

cmacq2:
I redid the import/EXTRA_COMPONENTS logic.

Now you can create an `include` file which works similar to package files and lists additional components to import line-by-line (one component per line).

Recursive imports are supported.

brickstrap-components.sh
```diff
@@ -18,6 +18,126 @@
...
+    fi
+
+    # Read include file
+    while IFS='' read -r BRP_CUR_LINE || [ -n "$BRP_CUR_LINE" ]; do
+        case "$BRP_CUR_LINE" in
+        \#*|\;*) # permit comments: lines starting with # or ; are ignored.
+        ;;
+        *)
+            # avoid redundant spaces, i.e.  empty lines are ignored.
+            # avoid adding duplicate componentn names
```
dlech: spelling "componentn"

brickstrap-components.sh
```diff
@@ -18,6 +18,126 @@
...
+            if [ -z "${BRP_CUR_LINE##/}" ] || \
```
dlech: should be `; then` instead of ` || \`

dlech:
Running 

    .../brickstrap.sh -p ev3dev-jessie -c ev3 -d ev3 create-rootfs

results in

    [ fail] At least one component is required

brickstrap-components.sh
```diff
@@ -18,6 +18,126 @@
...
+                brp_is_component_included "$BRP_CUR_LINE" || \
+                br_is_component_inherited "$BRP_CUR_LINE"
```
dlech: or should the `; then` be here?

brickstrap-components.sh
```diff
@@ -137,21 +258,46 @@ function brp_validate_component_names()
...
 #
+# $1: the list of components to iterate over.
+# $2: the callback to invoke.
+# $3...: optional: additional arguments passed to the callback following the
+#                  component name.
+#
+function brp_iterate_components_impl()
+{
+    # note: $2 may be empty, which is valid.
+    if [ $# -lt 3 -o -z "$1" -o -z "$3" ]; then
```
dlech: Looks like the hang-up is here. This fails if there are less than 3 arguments, but arg 3 is optional and `brp_iterate_components_impl` is called with 2 arguments from line 300.

brickstrap-components.sh
```diff
@@ -137,21 +258,46 @@ function brp_validate_component_names()
 # arguments passed to this function are passed along to the callback aftet the
```
dlech: spelling "aftet"

brickstrap-components.sh
```diff
@@ -137,21 +258,46 @@ function brp_validate_component_names()
...
+        done && return $BRP_COMP_CB_RETURNCODE
+    fi
+}
+
+
+#
+# Iterate over the configured component names (commandline arguments), invoking
+# callback for each component name. The calling convention is such that the
+# component name is passed as first argument to the callback, and any extra
+# arguments passed to this function are passed along to the callback aftet the
```
dlech: spelling "aftet"

cmacq2:
I've added a fix for missing/empty `/etc/shells` in the rootfs which causes `add-shell` from `debianutils` version 4.6 to fail. Without the fix `dash`, `bash` and other login shells aren't properly configured as part of `dpkg --configure -a`.

Additionally it appears that the 'run dpkg --configure twice' thing is no longer required so I removed it as well as the `|| true` bit.

cmacq2:
I've added another fix/improvement: 

 * The `gawk` package is technically an optional Debian package so brickstrap should not hard depend on the user install it as part of the multistrap. By contrast `mawk` is a required package, hence it should be available and can be used as fallback to provide a usable `awk`. 
 * Additionally it's possible to move the 'awk' symlink to the 'private' `/brickstrap` tree inside the chroot.

The latest update to the PR adds both of these improvements.

dlech:
This is too many unrelated changes for a single commit. Please split this up into one commit for each feature (and preferably make a separate pull request for each).

cmacq2:
Hmm, I suppose you are right. It started out a bit torn between "fix" and "feature" and the additional fixes don't make it any more focused.

... In the meantime, could you continue your review/testing of the actual code?

cmacq2:
I've created pull requests 41 thru 45 to isolate the bug fix/cleanup aspects of this PR.

The component import/inheriting feature has not been extracted yet: I expect it will create merge conflicts with the other PRs (due to spanning multiple files and offsets changing as a result of other PRs). 

It would be cleaner to merge/reject the fixes first and then rebase the import feature on top of the new master.

dlech:
The changes here seem to be working as advertised now.

cmacq2:
I've extracted the component import/inheritance feature and prepared it as a separate PR. As a result all changes have corresponding PRs of their own and so this PR can now be closed.

PR #40, opened 2016-01-23
Add new default image driver.

dlech:
This driver uses a single partition. Since brickstrap is mainly for debian
and most of the debian armmp platforms use a single partition, it makes
sense to have this as the default.


brickstrap-image-drivers.sh
```diff
@@ -22,6 +23,43 @@
...
+# Variables:
+# ROOT_PART_NAME: Label of root partition. Default: ROOTFS
+# IMAGE_FILE_SIZE: The size of the entire image file. Default: 3800M
+#
+function brp_image_drv_check_single_fs()
+{
+    ROOT_PART_NAME=${ROOT_PART_NAME:-ROOTFS}
+    IMAGE_FILE_SIZE=${IMAGE_FILE_SIZE:-3800M}
+
+    [ ${#ROOT_PART_NAME} -gt 16 ] && \
```
cmacq2: It might be preferable to extract this as utility function that takes two argument and returns either 0 or non-zero exit code. `brp_image_check_part_name` or something. Call like this:
```
brp_image_check_part_name ROOT_PART_NAME "$ROOT_PART_NAME"
```
... And update other fs drivers that want the same validation for partition names to take advantage of this as well?

cmacq2:
Did you add the second commit to this PR intentionally or was it meant to go to a different branch/PR...?

brickstrap-image-drivers.sh
dlech: Agreed.

dlech:
It's not exactly related, but I was being lazy since the second commit builds on the first and the second one is rather trivial.

dlech:
Added some more commits to extract common checks to utility functions and renamed some variables.

PR #45, opened 2016-01-27
Fixes to the package configure step in brickstrap

cmacq2:
* Add work around for empty/missing /etc/shells.
   The add-shell tool from debianutils package may choke on empty/missing /etc/shells, preventing packages like dash and bash from being configured properly.
 * Related: no longer hide errors from dpkg --configure -a as this is no longer necessary.
 * Do not hard depend on the user chosing to install gawk as part of the multistrap. The package is technically optional, and a fallback to mawk (which is a priority=required package) is possible.
 * Related: alter the 'awk' symlink/PATH workaround to have the dummy 'awk' symlink reside in the private /brickstrap/bin hierarchy within the chroot. (As opposed to /usr/local/bin).

As per request, this PR isolates and supersedes the corresponding change from PR #39

### DIFF
--- a/brickstrap.sh
+++ b/brickstrap.sh
@@ -428,15 +428,71 @@ function brp_preseed_debconf() {
     rm "$(br_rootfs_dir)/tmp/debconfseed.txt"
 }
 
+#
+# Ensure that a 'usable' /etc/shells file exists for the add-shell utility.
+# This ensures that configuration of shell packages (dash in particular) do
+# not fail with spurious error messages from misbehaving add-shell.
+# The add-shell utility of debianutils (at least version 4.6) may fail on
+# empty/missing /etc/shells file with a bogus error message.
+#
+function brp_fixup_etc_shells()
+{
+    info "Checking /etc/shells"
+    if [ ! -f "$(br_rootfs_dir)/etc/shells" ] ||
+        [ ! -s "$(br_rootfs_dir)/etc/shells" ]; then
+        if [ -f "$(br_rootfs_dir)/usr/share/debianutils/shells" ] &&
+            [ -s "$(br_rootfs_dir)/usr/share/debianutils/shells" ]; then
+            info "Populating default /etc/shells from template"
+            cp "$(br_rootfs_dir)/usr/share/debianutils/shells" \
+                "$(br_rootfs_dir)/etc/shells"
+        else
+            info "Generating dummy contents for: /etc/shells"
+            echo "# Dummy comment to work around add-shell bug" >\
+                "$(br_rootfs_dir)/etc/shells"
+        fi
+    else
+        info "Default /etc/shells appears to be sane... skipping"
+    fi
+}
+
 function brp_configure_packages () {
     info "Configuring packages..."
     brp_check_rootfs_dir
+    BRP_OLD_PATH="$PATH"
+    BRP_W_AWK_PATH="$PATH"
 
     # awk needs to be in the path, but Debian symlinks are not
-    # configured yet, so make a temporary one in /usr/local/bin.
-    info "Creating awk temporary symlink"
-    br_chroot mkdir -p /usr/local/bin
-    br_chroot ln -sf /usr/bin/gawk /usr/local/bin/awk
+    # configured yet, so make a temporary one in /$(br_chroot_brp_dir)/bin
+    # then export the dir on the PATH.
+    #
+    if br_chroot which awk >/dev/null; then
+        info "Using existing 'awk': $(br_chroot which awk)"
+    elif [ -x "/$(br_chroot_brp_dir)/bin/awk" ]; then
+        info "Reusing old temporary symlink: /$(br_chroot_brp_dir)/bin/awk"
+        BRP_W_AWK_PATH="$BRP_W_AWK_PATH:/$(br_chroot_brp_dir)/bin"
+    else
+        info "Creating awk temporary symlink: /$(br_chroot_brp_dir)/bin/awk"
+        br_chroot mkdir -p "/$(br_chroot_brp_dir)/bin"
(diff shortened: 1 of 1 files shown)


brickstrap.sh
```diff
@@ -428,15 +428,71 @@ function brp_preseed_debconf() {
     rm "$(br_rootfs_dir)/tmp/debconfseed.txt"
 }
 
+#
+# Ensure that a 'usable' /etc/shells file exists for the add-shell utility.
```
dlech: I have never had a problem with this in wheezy/jessie. Is this just a problem in testing/unstable?

brickstrap.sh
```diff
@@ -428,15 +428,71 @@ function brp_preseed_debconf() {
...
 
     # awk needs to be in the path, but Debian symlinks are not
-    # configured yet, so make a temporary one in /usr/local/bin.
-    info "Creating awk temporary symlink"
-    br_chroot mkdir -p /usr/local/bin
-    br_chroot ln -sf /usr/bin/gawk /usr/local/bin/awk
+    # configured yet, so make a temporary one in /$(br_chroot_brp_dir)/bin
+    # then export the dir on the PATH.
+    #
+    if br_chroot which awk >/dev/null; then
```
dlech: if `mawk` is a required package and `gawk` is not, lets just always use `mawk` and forget about `gawk`.

brickstrap.sh
```diff
@@ -461,14 +517,18 @@ function brp_configure_packages () {
...
-    br_chroot_bind /usr/bin/dpkg --configure -a || \
-    br_chroot_bind /usr/bin/dpkg --configure -a || true
+    br_chroot_bind /usr/bin/dpkg --configure -a
+
+    export PATH="$BRP_OLD_PATH"
 
     # remove our temporary awk symlink as it is no longer required.
-    info "Removing awk temporary symlink"
-    br_chroot rm -f /usr/local/bin/awk
+    if [ -x "/$(br_chroot_brp_dir)/bin/awk" ]; then
```
dlech: Since we are removing the symlink here, it seems like it would be simpler to just put it at `/usr/local/bin` instead of `/brickstrap/bin` so that we don't have to mess with the path variable here. If it were sticking around longer, it would make sense to have it under `/brickstrap` so that it was excluded from the tar file, but that is not the case here.

brickstrap.sh
```diff
@@ -461,14 +517,18 @@ function brp_configure_packages () {
         fi
     done
 
-    # run dpkg `--configure -a` twice because of errors during the first run
+    brp_fixup_etc_shells
+
     info "configuring packages..."
-    br_chroot_bind /usr/bin/dpkg --configure -a || \
-    br_chroot_bind /usr/bin/dpkg --configure -a || true
+    br_chroot_bind /usr/bin/dpkg --configure -a
```
dlech: I would prefer this change as a separate commit so that it can be reverted if needed. And actually the shell and awk stuff should be 2 separate commits as well.

brickstrap.sh
cmacq2: Yes, specifically debianutils 4.6.

It's because the maintainer of `debianutils` is looking for ways around using `awk` and his first attempt using `grep '^'` didn't work properly. I've also filed a bug about this here: https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=812784

In general I think it's a good idea for brickstrap to do this because `dash` is such a critical package. The alternative of potentially hiding failure to configure dash correctly is worse than having to carry a straightforward enough workaround.

brickstrap.sh
cmacq2: Well the original used `gawk` so I was a bit reluctant to change that. Is it really worth fixing on `mawk` instead of keeping the fallback mechanism as-is? 

In either case I'd keep the `if` guards that test if the `awk` symlink happens to be present already -- so it's really not a lot of code would be removed (just an `if` test + `ln -sf`.)

brickstrap.sh
dlech: This bug has already been fixed in unstable, so it seems like this workaround is no longer needed.

brickstrap.sh
cmacq2: I wanted it out of `/usr/local/bin` because `/usr` *should* not be messed with if at all possible. And this is quite possible.

That is precisely the reason we have the `/brickstrap`: for our temporary modifications to the rootfs and to hide those modifications from other code/tars/imaging where possible.

So I disagree: much better to locally redefine PATH for a short duration than to root around `/usr/local` which is for the configuration author to deploy their own apps to if these aren't package properly yet. (Following the somewhat dated principle of: /usr -> package manager; /usr/local -> sysadmin.)

brickstrap.sh
cmacq2: Hmm, it's kind of bound up with `/etc/shells` (in that: after the shells workaround this became possible).

brickstrap.sh
dlech: I had a look an `mawk` hasn't been touched since 2012, so yes, maybe better to keep `gawk` as well.

cmacq2:
I'll see about splitting this into 3 PRs.

dlech:
3 commits in 1 PR is OK too.

brickstrap.sh
dlech: This is something from the original polystrap script. It is there because there were a number of packages that were found to fail on the first go around. It is quite possible that this may be needed in the future.

cmacq2:
>. 3 commits in 1 PR is OK too.

Ah. I got the impression you preferred 3 PRs from the previous round of splitting :)

brickstrap.sh
dlech: OK.

cmacq2:
Splitting is done, I added the last `dpkg --configure -a` to the `/etc/shells/` workaround PR.

cmacq2:
Closing this PR because the changes in this branch that are not yet merged into master are now addressed in PR #48.

PR #48, opened 2016-01-27
Add work around for empty/missing /etc/shells.

cmacq2:
The add-shell tool from debianutils package may choke on empty/missing /etc/shells, preventing packages like dash and bash from being configured properly.

As per request, this PR isolates and supersedes the corresponding change from PR #45

### DIFF
--- a/brickstrap.sh
+++ b/brickstrap.sh
@@ -429,6 +429,33 @@ function brp_preseed_debconf() {
     rm "$(br_rootfs_dir)/tmp/debconfseed.txt"
 }
 
+#
+# Ensure that a 'usable' /etc/shells file exists for the add-shell utility.
+# This ensures that configuration of shell packages (dash in particular) do
+# not fail with spurious error messages from misbehaving add-shell.
+# The add-shell utility of debianutils (at least version 4.6) may fail on
+# empty/missing /etc/shells file with a bogus error message.
+#
+function brp_fixup_etc_shells()
+{
+    info "Checking /etc/shells"
+    if [ ! -f "$(br_rootfs_dir)/etc/shells" ] ||
+        [ ! -s "$(br_rootfs_dir)/etc/shells" ]; then
+        if [ -f "$(br_rootfs_dir)/usr/share/debianutils/shells" ] &&
+            [ -s "$(br_rootfs_dir)/usr/share/debianutils/shells" ]; then
+            info "Populating default /etc/shells from template"
+            cp "$(br_rootfs_dir)/usr/share/debianutils/shells" \
+                "$(br_rootfs_dir)/etc/shells"
+        else
+            warn "Generating dummy contents for: /etc/shells"
+            echo "# Dummy comment to work around add-shell bug" >\
+                "$(br_rootfs_dir)/etc/shells"
+        fi
+    else
+        info "Default /etc/shells appears to be sane... skipping"
+    fi
+}
+
 function brp_configure_packages () {
     info "Configuring packages..."
     brp_check_rootfs_dir
@@ -490,6 +517,8 @@ Tried: /usr/bin/mawk"
                     br_chroot_bind ${BRP_script##$(br_rootfs_dir)} install
         fi
     done
+
+    brp_fixup_etc_shells
 
     # run dpkg `--configure -a` twice because of errors during the first run
     info "configuring packages..."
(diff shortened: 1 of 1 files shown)


dlech:
Carrying on the conversation from <https://github.com/ev3dev/brickstrap/pull/45#discussion_r51057017>... I see that the bug has already been fixed in unstable, so it seems like there is no longer anything to work around.

cmacq2:
Yes, OTOH this workaround basically compensates for the fact that `brickstrap` does something different from what normal `apt-get install debianutils` is supposed to do.

Normal installation of `debianutils` is supposed to copy the `/usr/share/debianutils/shells` template to `/etc/shells` but for some reason this doesn't happen when using brickstrap. (You'll not that this is basically the first thing the workaround tries to do.)

So I'd keep the workaround because there is other 'funny' stuff going on here as well.

dlech:
> workaround basically compensates for the fact that brickstrap does something different from what normal apt-get install debianutils is supposed to do.

Actually, `debianutils` is working as intended. if `/etc/shells` exists already, it does not copy the template. <https://sources.debian.net/src/debianutils/4.7/debian/postinst/>

dlech:
Well, since `debianutils` is a required package, I guess I'm ok with leaving this in.

cmacq2:
> Actually, debianutils is working as intended

But not in our context: that was what stumped me before I figured out the work around:

 * `dash` depends on `debianutils`
 * `debianutils` is supposed to install a template `/etc/shells`
 * But somehow `add-shell` fails which can only happen if `/etc/shells` is either missing or empty. Since the template `/usr/share/debianutils/shells` is *not* empty that leaves...
 
Hence: workaround. Problem vanishes. And even with the update/revert to `debianutils` 4.7 I'd just leave this in as a safety catch/insurance against this stuff breaking in the future. (Because the knock-on effects of a broken `dash` or `bash` are no good.)

brickstrap.sh
```diff
@@ -429,6 +429,33 @@ function brp_preseed_debconf() {
...
+    info "Checking /etc/shells"
+    if [ ! -f "$(br_rootfs_dir)/etc/shells" ] ||
+        [ ! -s "$(br_rootfs_dir)/etc/shells" ]; then
+        if [ -f "$(br_rootfs_dir)/usr/share/debianutils/shells" ] &&
+            [ -s "$(br_rootfs_dir)/usr/share/debianutils/shells" ]; then
+            info "Populating default /etc/shells from template"
+            cp "$(br_rootfs_dir)/usr/share/debianutils/shells" \
+                "$(br_rootfs_dir)/etc/shells"
+        else
+            info "Generating dummy contents for: /etc/shells"
```
dlech: I think this should be a `warning` rather than info since it means that you are missing a required package.

dlech:
I wonder if a `preinst` script is touching `/etc/shells` and that is what is causing the problem. It's probably better to move this before the loop that runs the `preinst` scripts.

cmacq2:
How would that improve anything? Either the `preinst` scripts generate a non-empty `/etc/shells` in which case, fine (presumably) and the workaround is not triggered; or the scripts remove it/empty it in which case the workaround must be executed *after* the scripts run.

Either way after `preinst` seems the correct moment to invoke it.

brickstrap.sh
cmacq2: `info "Generating dummy contents for: /etc/shells"` you mean? I'd keep the other statements at `info` logging level.

dlech:
> Either the preinst scripts generate a non-empty /etc/shells in which case, fine (presumably) and the workaround is not triggered;

If a `preinst` script generates a non-empty file, then `/bin/sh` will be missing from `/etc/shells` because [that comes from the template.](https://sources.debian.net/src/debianutils/4.7/debian/shells/)

cmacq2:
> If a preinst script generates a non-empty file, then /bin/sh will be missing from /etc/shells because that comes from the template.

As I understand it: `/bin/sh` will be added during `dpkg --configure -a` when `dash` is configured and the `dash` package invokes the `add-shell` tool to register itself in `/etc/shells`.

cmacq2:
By the way I've rebased the PR against master and fixed up the info/warning message thing.

cmacq2:
Is this good to go or does it require further discussion?

dlech:
My gut feeling tells me that just copying `/usr/share/debianutils/shells` to `/etc/shells` before running the preinst scripts is the "right thing" to do but I am tired of arguing about it. The PR is good enough as-is and I will merge when I have a chance to run through an image build again.

PR #50, opened 2016-01-29
Introduce a 'reports' directory

cmacq2:
Introduce a 'reports' directory to aggregate reports generated via custom-report.sh

With this change custom-report.sh instances will have a well-defined location to put their reports (example: `$(br_report_dir)/your_report_name_here.txt`), and brickstrap has a well-defined location to delete, namely `$(br_report_dir)` itself.

Introduce the `br_output_basename` function so reporting scripts have a well defined template basename for report file names.


cmacq2:
Any thoughts on this?

dlech:
Looks good to me. I might be inclined to call the `br_output_basename` function `br_image_basename` instead.

cmacq2:
I thought of that but decided against it when I wrote the code because:
 * It is intended for use by scripts such as `custom-report.sh` for filenames of arbitrary output artifacts
 * There's also `brp_image_name` which reflects the actual image base name. It seemed a good idea to preserve the distinction/avoid confusion in the function name.

dlech:
Isn't `brp_image_name` the full name (including file extension)? Also, since the "output" name can be specified with the `-I` option (advertised as image base name), I think it makes more sense to use `br_image_basename` unless you are planning on making it more generic by updating the docs and maybe changing the parameter to `-o`.

cmacq2:
> Isn't brp_image_name the full name (including file extension)?

Correct.

> I think it makes more sense to use br_image_basename

Done, in the new version of this PR.

PR #53, opened 2016-02-08
Make the standalone copy-root command more robust.

cmacq2:
With this change, brickstrap will create the rootfs directory if necessary (when it does not exist yet) during a standalone copy-root command.
Previously a missing rootfs directory would trigger a fatal error during brp_copy_root().


dlech:
LGTM. Thanks.

PR #56, opened 2016-06-11
Commented entry for BrickPi Input Port 5 I2C baudrate

bmegli:
Hmm, now I will also have to update the docs on github.io to uncomment
the line instead of adding it...

### DIFF
--- a/projects/ev3dev-jessie/rpi-base/root/boot/flash/config.txt
+++ b/projects/ev3dev-jessie/rpi-base/root/boot/flash/config.txt
@@ -27,6 +27,10 @@
 # by default the Bluetooth uses the same UART as BrickPi
 #dtoverlay=pi3-disable-bt
 
+# uncomment only if you are using BrickPi (not BrickPi+) 
+# and want to use Input Port 5 for I2C/NXT sensor 
+#dtparam=i2c_baudrate=9600
+
 # uncomment to disable BrickPi LEDs on boot.
 #dtparam=brickpi_led1_trigger=none,brickpi_led2_trigger=none
 
--- a/projects/ev3dev-jessie/rpi-base/root/boot/flash/config.txt
+++ b/projects/ev3dev-jessie/rpi-base/root/boot/flash/config.txt
@@ -29,6 +29,7 @@
 
 # uncomment only if you are using BrickPi (not BrickPi+) 
 # and want to use Input Port 5 for I2C/NXT sensor, some sensors
+# work at higher baudrates so you may want to tweak the value 
 #dtparam=i2c_baudrate=9600
 
 # uncomment to disable BrickPi LEDs on boot.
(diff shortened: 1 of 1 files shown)


projects/ev3dev-jessie/rpi-base/root/boot/flash/config.txt
```diff
@@ -27,6 +27,10 @@
 # by default the Bluetooth uses the same UART as BrickPi
 #dtoverlay=pi3-disable-bt
 
+# uncomment only if you are using BrickPi (not BrickPi+) 
+# and want to use Input Port 5 for I2C/NXT sensor
```
dlech: We could add a sentence to say that some sensors work at higher baud rates

projects/ev3dev-jessie/rpi-base/root/boot/flash/config.txt
```diff
@@ -27,6 +27,10 @@
 # by default the Bluetooth uses the same UART as BrickPi
 #dtoverlay=pi3-disable-bt
 
+# uncomment only if you are using BrickPi (not BrickPi+) 
+# and want to use Input Port 5 for I2C/NXT sensor 
+#dtparam=i2c_baudrate=9600
```
dlech: EV3 actually uses `10000` instead of `9600`

dlech:
FYI, these changes won't ship until we do a new image release.

projects/ev3dev-jessie/rpi-base/root/boot/flash/config.txt
bmegli: Yes, you are right in both. Probably it's better to run at higher if possible.
I only have CruizCore little to test here, it works ok at 9600 and higher, but starting from some baudrate it stops working properly (no errors in dmesg but -1 on angle value for example all the time)

Mindstorms NXT Hardware Developper Kit (page 9) states that: 
> I2C communication is running at 9600 bit/s

So at least in theory I2C/NXT sensors should at least work at this (slow!) speed and were tested for it.

But yes, I will update the docs and mention the somebody may want to tweak the value.

bmegli: Do you think we should use `10000` as in EV3 as the default value?

dlech: Yes. it is at least a *tiny* bit faster.

projects/ev3dev-jessie/rpi-base/root/boot/flash/config.txt
```diff
@@ -28,7 +28,7 @@
 #dtoverlay=pi3-disable-bt
 
 # uncomment only if you are using BrickPi (not BrickPi+) 
-# and want to use Input Port 5 for I2C/NXT sensor 
+# and want to use Input Port 5 for I2C/NXT sensor, some sensors
```
dlech: I think you missed part of the sentence.

bmegli:
Thank you. I suppose I have not hit ctrl+s in text editor before commit

Issue #60, opened 2016-09-20
brickstrap tar error


vladshablinsky:
Hi, I'm trying to run `brickstrap` on my docker image I imported to docker, but it fails.
I had an image on my OS X machine and exported it to tar into shared folder and then imported it from ubuntu VM into docker. Now I'm getting the following error while running `create-tar`.

I tried both running brickstrap as a shell script and installed using `apt`.

```
➜ viki:git$ ./brickstrap/brickstrap.sh create-tar b01ab378b678 OUTPUT.tar
Checking docker image tar version...
tar 1.27.1-2+b1
Creating OUTPUT.tar from b01ab378b678...
/bin/tar: /brickstrap/_tar-exclude: No such file or directory
/bin/tar: Error is not recoverable: exiting now
```

Here are my tar version and docker on the guest.
```
➜ viki:git$ tar --version
tar (GNU tar) 1.27.1
Copyright (C) 2013 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>.
This is free software: you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.

Written by John Gilmore and Jay Fenlason.
➜ viki:git$ docker -v
Docker version 1.12.1, build 23cf638
```

Here is my docker on the host.

```
➜ vlad:git$ docker --version
Docker version 1.12.1, build 6f9534c
```

What should I do? Thanks!

dlech:
You should send a pull request that checks if the file `/brickstrap/_tar-exclude` exists in the docker image. The check should go [here](https://github.com/ev3dev/brickstrap/blob/master/brickstrap.sh#L66) and only set `BRICKSTRAP_TAR_EXCLUDE_OPTION` if the file exists.

The check might look something like this:

    if docker run --rm $BRICKSTRAP_DOCKER_IMAGE_NAME test -f /brickstrap/_tar-exclude; then
        BRICKSTRAP_TAR_EXCLUDE_OPTION="--exclude-from /brickstrap/_tar-exclude"
    fi

dlech:
But you probably really want to exclude some files, like `/usr/sbin/policy-rc.d`.

vladshablinsky:
@dlech I've opened a PR as you suggested. Btw, It fixes the error, which is not so much of surprise, though. :)

> But you probably really want to exclude some files, like /usr/sbin/policy-rc.d.

[...]

PR #61, opened 2016-09-20
Check if /brickstrap/_tar-exclude exists in docker

vladshablinsky:
Brickstrap fails if we don't check that `/brickstrap/_tar-exclude` exists.

```
➜ viki:git$ ./brickstrap/brickstrap.sh create-tar b01ab378b678 OUTPUT.tar
Checking docker image tar version...
tar 1.27.1-2+b1
Creating OUTPUT.tar from b01ab378b678...
/bin/tar: /brickstrap/_tar-exclude: No such file or directory
/bin/tar: Error is not recoverable: exiting now
```

See #60.

### DIFF
--- a/brickstrap.sh
+++ b/brickstrap.sh
@@ -63,9 +63,10 @@ function brickstrap_create_tar()
     if dpkg --compare-versions $BRICKSTRAP_DOCKER_IMAGE_TAR_VERSION ge 1.28; then
         BRICKSTRAP_TAR_EXCLUDE_OPTION="--exclude-ignore .brickstrap-tar-exclude"
     else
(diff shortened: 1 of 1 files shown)


dlech:
Have you tested this to make sure it works?

Also, please indent using 4 spaces. Other than that, it looks good to me.

vladshablinsky:
> Have you tested this to make sure it works?

Well, I tested it on a container with and without `/brickstrap/_tar-exclude` file and both of them finished with `done`.

> Also, please indent using 4 spaces.

Sorry for that. I usually use two spaces and didn't noticed that four spaces are used here.

dlech:
Excellent! Thank you very much.

PR #64, opened 2017-08-30
Use built-in tar version command instead of dpkg

hashworks:
Since dpkg isn't available on non-debian systems it's better to use the build in `tar --version` command to check the tar version. Comparisation happens with the version-sort feature from coreutils `sort`.

Things I have tested:
* `BRICKSTRAP_DOCKER_IMAGE_TAR_VERSION` output
* `SORTED_VERSION` output
* Setting `BRICKSTRAP_DOCKER_IMAGE_TAR_VERSION=1.27`
* Setting `BRICKSTRAP_DOCKER_IMAGE_TAR_VERSION=1.28`

### DIFF
--- a/src/brickstrap.sh
+++ b/src/brickstrap.sh
@@ -51,17 +51,17 @@ function brickstrap_create_tar()
         exit 1
     fi
 
-    # --exclude-ignore requires tar >= 1.28, so first we check the tar version in
+    # --exclude-ignore requires tar > 1.27, so first we check the tar version in
     # the docker image.
 
     echo "Checking docker image tar version..."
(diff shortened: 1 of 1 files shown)


hashworks:
54216dd failed because I forgot tar version `1.27.1`. Commit f0fd65c fixed that. According to the [offical download folder](http://ftp.gnu.org/gnu/tar/) this is the last version before `1.28`.

hashworks:
Actually... yes. I don't know why I didn't do it like this right away.

dlech:
Looks great. Thanks!

